Issue #227, opened 2026-01-16
Integrate `check-spelling` into CI to catch documentation typos automatically


Ryan-Millard:
> [!TIP]
> This issue hasn't been claimed yet. Comment `/take` if you'd like to work on it!

---

## Description

Img2Num has a growing amount of user-facing documentation (README, Docusaurus docs, guides, comments, etc.). To improve documentation quality and reduce review overhead, we should integrate [`check-spelling`](https://github.com/check-spelling/check-spelling) into our CI.

`check-spelling` is a GitHub Action specifically designed for open source projects and supports:

* Markdown, reStructuredText, plain text, and source comments
* Project-specific allowlists / wordlists
* Incremental adoption (only fail on new issues)
* Very low maintenance once configured

## Why this is useful

* Prevents repeated typo-related review comments
* Improves perceived project quality and approachability
* Encourages contributors to write cleaner docs
* Widely used and battle-tested in OSS projects

## Proposed scope

* Run on PRs and pushes to `main`
* Check:

  * `README.md`
  * `CONTRIBUTING.md`
  * `docs/**` (Docusaurus content)
  * Possibly comments in source files (configurable)
* Start with a permissive configuration (we'll tighten it later if necessary)

## Implementation sketch

* Add a workflow under `.github/workflows/check-spelling.yml`
* Include a `.github/.check-spelling/` directory for:

  * `allow.txt` (project-specific terms like Img2Num, SVG, WebGPU, etc.)
  * optional configuration files

## Example
The upstream repo provides a minimal starter workflow that can be adapted almost verbatim:
[https://github.com/check-spelling/check-spelling#example-workflow](https://github.com/check-spelling/check-spelling#example-workflow)

**Notes**

* This does *not* block contributors with false positives long-term, as words can be added to the allowlist.
* We can initially configure it to comment warnings instead of failing the build, if preferred.

---

# Something to get you started

## GitHub Actions workflow

Create this file:

**`.github/workflows/check-spelling.yml`**

```yaml
name: Check spelling

on:
  pull_request:
  push:
    branches:
      - main

permissions:
  contents: read
  pull-requests: write

jobs:
  spelling:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: Check spelling
        uses: check-spelling/check-spelling@v0.0.24
        with:
          suppress_push_for_open_pull_request: true
          post_comment: true
          fail_on_error: false
```

### Why this config

[...]

Issue #214, opened 2026-01-08
refactor: Streamline Documentation for Easier Onboarding


Ryan-Millard:
> [!TIP]
> This issue hasn't been claimed yet. Comment `/take` if you'd like to work on it!

---

## Problem

Currently, the [Guidelines](https://ryan-millard.github.io/Img2Num/info/docs/category/guidelines/) docs are helpful, but the overall structure can be overwhelming for new developers. In particular, the [Getting Started](https://ryan-millard.github.io/Img2Num/info/docs/introduction/getting-started/) page is too complex and difficult to follow, making onboarding slower and more confusing.

## Proposed improvements

* Refactor the Getting Started guide to be simpler, more concise, and more structured.
* Break complex sections into smaller, step-by-step instructions.
* Include clear examples where possible to illustrate key concepts.
* Consider linking back to the Guidelines only where absolutely necessary to avoid overwhelming new developers.
* Review the overall navigation flow so new contributors can quickly find what they need without feeling lost.

## Goal

Make the documentation more beginner-friendly, reduce onboarding friction, and help new developers start contributing to Img2Num more confidently.

## Guide to help you get started

1. Split the [Getting Started](https://ryan-millard.github.io/Img2Num/info/docs/introduction/getting-started/) page into a folder with a few pages inside that direct the user to the correct place. This will help Algolia index that page as well.
2. Include links to the relevant pages in the [Guidelines](https://ryan-millard.github.io/Img2Num/info/docs/category/guidelines/) pages and improve the content of each page. Include:
    - Helpful links (e.g., a YouTube video that explains how to use PR templates)
    - Extensive explanations of how things work (such as claiming issues)

> [!IMPORTANT]
> Ensure all documentation is up-to-date.

laxitajain:
> Thanks again to [@laxitajain](https://github.com/laxitajain) for the lovely issue claim system refactor!

The pleasure's all mine! :)

PR #245, opened 2026-01-27
feat(svg-builder): build SVGs after Suzuki-Abe topology extraction

Krasner:
## What was changed & why

Reworks SVG contour export. With the Graph and Node structure, currently contours are computed per node using the Suzuki-Abe method. While these contours are correct, neighboring regions will have non-overlapping contours - they will be off by 1 pixel. For SVG this is a problem - neighboring contours will show up doubled up.

We need a method to merge neighboring contours into one line.

Fixes:

## Changes

1. Move contour management to Graph - it will call nodes to compute their own contours, then will manage merging.
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/kmeans_graph.cpp#L221
2. After the nodes compute their contours, they are laid out on a binary image where the double borders will be apparent. Use skeletonization to thin all borders to 1 pixel thickness
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/graph.cpp#L165

3. For each node, we must adjust its contour to match these thinned boundaries. We have to determine which contour pixels got removed and/or shifted, and so we must search for the nearest skeletonized pixel
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/graph.cpp#L179

3a. To do this we take the binary representation of each Node (`n->create_binary_image`) and we figure out how to add more pixels to its boundary so that the resultant contour will fall along the skeletonized borders.
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/graph.cpp#L261

3b. once the node binaries are updated, we recompute contours - these should be VERY close to the skeletonized borders.
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/graph.cpp#L384

What we have done at this point is force neighboring regions to produce contour lines that overlap on the same pixels. For SVG this will remove the double border problem.

4. However some _holes_ still exist! We correct this as a packing problem. Envision all regions are in a 2D box. We want to maximally inflate each region to tightly press along neighboring regions. We can additionally apply contour smoothing here. To prevent holes, neighboring contours MUST follow each other EXACTLY!
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/graph.cpp#L424

5. We update node contours with these new smoothed and packed contours
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/graph.cpp#L435

6. We export these contours to the SVG exporter
https://github.com/Krasner/Img2Num/blob/41ce003750ce8902473ec547df3b577754a3c30f/src/wasm/modules/image/src/kmeans_graph.cpp#L224


Ryan-Millard:
> packWithBoundaryConstraints is only there to help reduce holes (and optionally smooth contours)
> 
> Skeletonization is a global operation. I take each node contour and place it on a grid (like you saw in cow_contours.txt). Skeletonization then thins the boundaries - it can tell me which pixels got removed but it does not tell me where to move them. Then I iterate over each node and have to figure out how to adjust each contour to match the skeletonized/thinned boundaries. That is the big for loop that searches for where might the contour point move if it got removed in the skel. step.
> 
> Theoretically that shouldn't leave any holes... but for some reason it still does. (occasional small holes). packWithBoundaryConstraints is supposed to fix that - but i don't think it's a good long term solution. Rather I need to figure out why the adjusted contours have those little holes.

I'll have a look. I'm taking a bit of strain because I'm suffering from burnout and realized that I need to take more time to touch grass because I used every spare second I had on this project previously.😂 I'll try to see what about helping, but I don't know how far I'll get.

I think `packWithBoundaryConstraints` is a good idea, but shouldn't be applied to the entire image because it distorts spatial data terribly. Instead, I think we need to find a way to detect whether skeletonization left any holes and then only apply `packWithBoundaryConstraints` to the hole's borders.

Ryan-Millard:
Would you be comfortable with reviewing a PR that is purely React? You can say no - I don't mind.

I sometimes spend a lot of time on React PRs and it would be nice to share the load with someone.

src/wasm/modules/image/src/contours.cpp
```diff
@@ -285,4 +289,631 @@ ContoursResult find_contours(const std::vector<uint8_t> &binary, int width,
...
+  return normalize({vec[next].x - vec[prev].x, vec[next].y - vec[prev].y});
+}
+
+float distSq(Point a, Point b) {
+  return (a.x - b.x) * (a.x - b.x) + (a.y - b.y) * (a.y - b.y);
+}
+// Calculate the closest point on the segment V -> W from point P
+// Returns {ClosestPoint, DistanceSquared}
+std::pair<Point, float> getClosestPointOnSegment(Point p, Point v, Point w) {
+  float l2 = distSq(v, w);
```
Ryan-Millard: Why is it called l2? I don't really understand this code and there is no documentation for it, so it might be a bit better to use clearer names.

Ryan-Millard:
Hi @Krasner, I’ve spent some time reading through this PR, but I’m still struggling to follow the code. The concepts involved here are quite advanced, and I don’t have enough background in these specific topics to confidently reason about the correctness or long-term maintainability of the code as it stands.

Before this can be merged, I’d really like to see some documentation added that explains:
- The problem this code solves
- The core concepts & algorithms involved (at a high level)
- Links to good external resources or papers for people who want to learn more

It doesn’t need to be as extensive as the FFT documentation
([https://ryan-millard.github.io/Img2Num/info/docs/reference/wasm/modules/image/fft_iterative/](https://ryan-millard.github.io/Img2Num/info/docs/reference/wasm/modules/image/fft_iterative/)),
but it should be enough to help someone with no prior background understand what the code is doing and why. The [Prerequisite Theory section](https://ryan-millard.github.io/Img2Num/info/docs/category/prerequisite-theory-for-ffts/) shows this quite well by listing things you need to know before learning about FFTs and by explaining how FFTs work. 

Documentation is especially important for this repository because we accept external contributions. While it’s true that people unfamiliar with these concepts shouldn’t modify this code directly, a lack of documentation raises the barrier unnecessarily and prevents contributors from ever getting to that point.

To me, Img2Num is more than just an open-source project - it’s also a learning space that should welcome newcomers. Clear documentation is a big part of making that possible.

Your code looks good, but I unfortunately need to understand it better before I can confirm that and it might take a while longer.

Krasner:
I think the most important concept in this repo is the Graph/Node data structure - it's used pretty extensively for managing regions in the image.

Give me some time to write this up... (as you can tell documentation is not my strength...)

Krasner:
@Ryan-Millard had to implement the bezier stuff... It reduces the SVG string size so i can render the big image SVGs now too 
image
ok now will work on documentation...

docs/docs/reference/wasm/modules/image/graph/explained.md
```diff
@@ -0,0 +1,78 @@
...
+sidebar_position: 3
+---
+
+To explain Graph creation from image regions we show a diagram representation followed by a concrete step-by-step example.
+
+## Diagram
+Assume a region partition image as shown. Each partition (and the pixels contained within) are represented as a Node. Nodes exist in the heap. A Graph is a collection of shared_ptr(s) to these Nodes. A Graph has unique ownership over these Nodes pointers.
+![slide1](./diagrams/Slide1.SVG)
+
+This is how one would imagine the Graph to be connected forming a undirected graph(a). In reality each Node contains a list of pointer to their neighbors, represented as edges (b). Note that adjacent nodes point to each other. Edge management is handled by the Graph.
```
Ryan-Millard:
```suggestion
Below is a visualization of the Graph with its connections forming a undirected graph (a). In reality each `Node` contains a list of pointers to their neighbors, represented as edges (b). Note that adjacent nodes point to each other. Edge management is handled by the Graph.
```

docs/docs/reference/wasm/modules/image/graph/explained.md
Ryan-Millard: Please add captions to all of your figures like the below to make referencing them simpler:

```md
![Alt text for the image](./img/my-image.png)

*Figure 1: This is the caption.*
```

docs/docs/reference/wasm/modules/image/graph/explained.md
```diff
@@ -0,0 +1,78 @@
...
+id: explained
+title: Graph Explained
+sidebar_position: 3
+---
+
+To explain Graph creation from image regions we show a diagram representation followed by a concrete step-by-step example.
+
+## Diagram
+Assume a region partition image as shown. Each partition (and the pixels contained within) are represented as a Node. Nodes exist in the heap. A Graph is a collection of shared_ptr(s) to these Nodes. A Graph has unique ownership over these Nodes pointers.
+![slide1](./diagrams/Slide1.SVG)
```
Ryan-Millard:
```suggestion
![slide1](./diagrams/Slide1.SVG)
*Figure 1: Mapping regions in an image (likely quantized) into a Graph's nodes.
```

Ryan-Millard:
```suggestion
## Mapping Images to Graphs
Assume a region partition image as shown. Each partition (and the pixels contained within) are represented as a Node. Nodes exist in the heap. A Graph is a collection of shared pointers to these Nodes (`std::unique_ptr<std::vector<std::shared_ptr<Node>>>`). A Graph has unique ownership over these Nodes pointers.
![slide1](./diagrams/Slide1.SVG)
*Figure 1: Mapping regions in an image (likely quantized) into a Graph's nodes.
```
```

File: docs/docs/reference/wasm/modules/image/graph/explained.md [hunk 4]
```diff
@@ -0,0 +1,78 @@
...
+2. Building graphs consists of 2 steps:
+    - Uniquely label neighboring regions. Floodfill is used for this. This discovers connected components - each connected component becomes a `Node`
+
+    ![Regions](./img/regions.png)
+
+    - Tracking neighbors - the region labels are parsed to check for neighbors by looking at 3x3 neighborhoods. Each `Node` is updated with a list of neighbors forming an undirected graph.
+
+    ![Graph](./img/region_graph.png)
+
+NOTE: Since many regions are concave, their centroids may appear to be lying far away, but this is correct. Centroids here are just used for visualizing the graph.
```
Ryan-Millard:
```suggestion
    - Tracking neighbors - the region labels are parsed to check for neighbors by looking at 3x3 neighborhoods (Moore's 8 Neighbors). Each `Node` is updated with a list of neighbors forming an undirected graph as shown below (each centroid is plotted).

    ![Graph](./img/region_graph.png)

:::note Unexpected Centroid Locations
Since many regions are concave, their centroids may appear to be lying far away, but this is correct. Centroids here are just used for visualizing the graph.
:::
```

For this image, is it necessary that lines are drawn between centroid plots? Also, could they be given unique colors?

docs/docs/reference/wasm/modules/image/graph/explained.md
```diff
@@ -0,0 +1,78 @@
...
+
+3. Merging small regions
+
+The `Graph` is parsed to locate small area nodes. A small nodes are merged with a neighboring node by transfering its pixels and edges to the absorbing node.
```
Ryan-Millard:
```suggestion
The `Graph` is parsed to locate small area nodes. Each small node is merged with a larger neighbor node by transferring its pixels and edges to the absorbing neighbor node.
```

docs/docs/reference/wasm/modules/image/graph/explained.md
```diff
@@ -0,0 +1,78 @@
...
+
+4. Contour creation
+
+Now that small nodes have been pruned, each `Node` can compute its own contour using the Suzuki-Abe method.
+
+![Contours](./img/contours.png)
+
+However, while correct adjacent nodes will have parallel contours which becomes a problem in SVG creation.
+
+![ContoursZ](./img/contours_zoom.png)
```
Ryan-Millard:
```suggestion
:::caution SVG space offset
While Suzuki-Abe contour tracing correctly captures topological data, adjacent nodes end up with contours offset by ~1 pixel when mapped to SVG coordinates.

<div align="center">
See [this PR discussion](https://github.com/Ryan-Millard/Img2Num/pull/245#issuecomment-3807553622) for details on why this happens.
![ContoursZ](./img/contours_zoom.png)
</div>
:::
```

docs/docs/reference/wasm/modules/image/graph/api.md
```diff
@@ -0,0 +1,96 @@
...
+`Node`s reference neigbors through node pointers (`shared_ptr`)
+```
+Node_ptr n_ptr = std::make_shared<Node>(<id>, <std::unique_ptr<std::vector<RGBXY>> pixels>);
+```
+
+A `Graph` takes ownership over a collection of Nodes. It does so by referencing a list of Node pointers.
+```
+std::unique_ptr<std::vector<Node_ptr>> node_ptr =
+      std::make_unique<std::vector<Node_ptr>>(std::move(nodes));
+Graph G(node_ptr, width, height);
```
Ryan-Millard:
```suggestion
```cpp title="A Graph takes ownership over a collection of Nodes. It does so by referencing a list of Node pointers."
std::unique_ptr<std::vector<Node_ptr>> node_ptr = std::make_unique<std::vector<Node_ptr>> std::move(nodes));
Graph G(node_ptr, width, height);
```

docs/docs/reference/wasm/modules/image/graph/api.md
```diff
@@ -0,0 +1,96 @@
...
+
+In sum, `Graph`s manage a list of Nodes through their pointers. Each `Node` can reference neighboring nodes as edges also through their pointers.
+
+Since multiple entities can reference the same `Node` we use `shared_ptr`.
+
+# Usage
+This follow the step-by-step guide in the [explanation](explained.md).
+
+1. Graph creation from kmeans labels
+- Initialze nodes and region map
```
Ryan-Millard:
```suggestion
This follows the step-by-step guide in the [explanation](explained.md).

1. Graph creation from kmeans labels
- Initialize nodes and region map
```

We really need #227 to be completed.😅

docs/docs/reference/wasm/modules/image/graph/api.md
```diff
@@ -0,0 +1,96 @@
...
+}
+
+for node in G.nodes
+{
+    // compute contour per node
+}
+```
+
+4. Collect all contours for SVG export
+![code2](../../../../../../../src/wasm/modules/image/src/kmeans_graph.cpp#L252)
```
Ryan-Millard: Did these C++ file links work for you?

Ryan-Millard:
> sounds great! sorry this PR so big. It covers a lot of ground though!

It's a really solid PR, so that's not a problem. The docs have really helped me understand what's going on and you explained it well. 

Please would you add some docs for the skeletonization, suzuki-abe, and other important algorithms. It doesn't need to be much - just an explanation of how they work at a high level. 

I believe Jupyter Notebooks allow conversion to markdown, so you could use that to help you get started if you have any.

src/wasm/modules/image/include/bezier.h
```diff
@@ -0,0 +1,19 @@
...
+#include <cmath>
+#include <limits>
+#include <vector>
+
+// defined in contours.h
+/*struct QuadBezier {
+    Point p0; // Start
+    Point p1; // Control
+    Point p2; // End
+};*/
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/include/graph.h
```diff
@@ -31,14 +33,16 @@ discover_edges(G, region_labels, width, height);
 
 class Graph {
 protected:
+  int m_width, m_height;
```
Ryan-Millard: What is the `m_` prefix about in this class? I've never understood it.

Additionally, are these the dimensions of the image it maps over?

src/wasm/modules/image/include/node.h
```diff
@@ -35,6 +36,9 @@ discover_edges(G, region_labels, width, height);
 
 struct XY {
   int32_t x, y;
+  std::pair<int32_t, int32_t> xy;
+  XY(int32_t x_, int32_t y_) : x(x_), y(y_) { xy = std::make_pair(x, y); };
+  bool operator<(const XY &rhs) const { return xy < rhs.xy; };
 };
```
Ryan-Millard: We already have XY in another file, so it might be better to make a new file that we can include that contains this information.

src/wasm/modules/image/include/node.h
```diff
@@ -68,15 +76,23 @@ class Node {
...
+  ColoredContours m_contours;
+  void compute_contour();
+
   /* access member variables */
   inline int32_t id() const { return m_id; };
   inline size_t area() const { return m_pixels->size(); };
   inline const std::set<Node_ptr> &edges() const { return m_edges; }
   inline size_t num_edges() const { return m_edges.size(); }
   inline const std::vector<RGBXY> &get_pixels() const { return *m_pixels; }
+  inline ColoredContours &get_contours() { return m_contours; }
```
Ryan-Millard: Should `m_contours` be mutable from outside the class?

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
+#include "bezier.h"
+#include <iostream>
+
+// --- Vector Math Helpers ---
+inline Point sub(Point a, Point b) { return {a.x - b.x, a.y - b.y}; }
+inline Point add(Point a, Point b) { return {a.x + b.x, a.y + b.y}; }
+inline Point mul(Point a, float s) { return {a.x * s, a.y * s}; }
+inline float dot(Point a, Point b) { return a.x * b.x + a.y * b.y; }
```
Ryan-Millard: I think operator overloading makes more sense for this instead.

`distSq` could also be a member (maybe `inline float distSqFrom(Point other)`), but that's optional.

Krasner:
Yes you're trying to fit to this polynomial
```
 P(t) = (1-t)^2 Q0 + 2t(1-t) Q1 + t^2 Q2
```
you need at least 3 points

Krasner:
I think the Bezier stuff needs a bit of a rework for the next PR. They don't always look as smooth as i would like them too. but it does help massively in reducing the SVG string size

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
...
+inline float distSq(Point a, Point b) {
+  return (a.x - b.x) * (a.x - b.x) + (a.y - b.y) * (a.y - b.y);
+}
+inline float len(Point a) { return std::sqrt(a.x * a.x + a.y * a.y); }
+
+// --- Evaluate Quadratic Bezier at t ---
+Point evalBezier(const QuadBezier &b, float t) {
+  float invT = 1.0 - t;
+  // B(t) = (1-t)^2 * P0 + 2*t*(1-t) * P1 + t^2 * P2
```
Ryan-Millard:
```suggestion
// B(t) = (1-t)^2 * P0 + 2*t*(1-t) * P1 + t^2 * P2
Point evalBezier(const QuadBezier &b, float t) {
  float invT = 1.0 - t;
```

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
...
+  float c1 = 2.0f * t * invT;
+  float c2 = t * t;
+  return {c0 * b.p0.x + c1 * b.p1.x + c2 * b.p2.x,
+          c0 * b.p0.y + c1 * b.p1.y + c2 * b.p2.y};
+}
+
+// --- Chord Length Parameterization ---
+// Assigns a 't' value (0.0 to 1.0) to each point based on distance
+std::vector<float> chordLengthParameterize(const std::vector<Point> &points,
+                                           int first, int last) {
```
Ryan-Millard: Why not expect that the caller of the function has already ensured the first element is the first we want to deal with and the last element is the last we want to deal with? I find it a bit strange to pass more data to a functoin than it needs because there is always a risk of accidental mutation.

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
...
+  float c0 = invT * invT;
+  float c1 = 2.0f * t * invT;
+  float c2 = t * t;
+  return {c0 * b.p0.x + c1 * b.p1.x + c2 * b.p2.x,
+          c0 * b.p0.y + c1 * b.p1.y + c2 * b.p2.y};
+}
+
+// --- Chord Length Parameterization ---
+// Assigns a 't' value (0.0 to 1.0) to each point based on distance
+std::vector<float> chordLengthParameterize(const std::vector<Point> &points,
```
Ryan-Millard: If all of the points are equidistant, **uniform parameterization** may be a bit better here.

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// --- Least Squares Fit for Control Point Q1 ---
+// We know Q0 (Start) and Q2 (End). We need to find Q1 that minimizes error.
+// Based on equation: P(t) = (1-t)^2 Q0 + 2t(1-t) Q1 + t^2 Q2
+// Rearranged: Q1 * [2t(1-t)] = P(t) - (1-t)^2 Q0 - t^2 Q2
+Point generateQuadBezier(const std::vector<Point> &points, int first, int last,
+                         const std::vector<float> &u) {
+  Point Q0 = points[first];
+  Point Q2 = points[last];
```
Ryan-Millard: I don't think this pattern is good from an architectural standpoint. It should be the caller's job to prepare the data for the function, not the other way around. I already mentioned the possibility of accidental mutation, but it's more than that.

One wouldn't hand their entire schoolbag to their teacher when handing in their homework, so I don't think it would be right to pass entire vectors to these functions. It isn't that it's a terrible decision, but more that the operation is unrelated to the function.

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
...
+    // Fit is bad, split at the point of maximum error
+    // Important: Prevent infinite recursion if split doesn't advance
+    if (splitPoint == first || splitPoint == last) {
+      // Fallback: simply bisect indices if geometric split fails
+      splitPoint = (first + last) / 2;
+    }
+
+    fitRecursive(points, first, splitPoint, errorLimit, outCurves);
+    fitRecursive(points, splitPoint, last, errorLimit, outCurves);
+  }
```
Ryan-Millard:
```suggestion
  if (maxDistSq < (errorLimit * errorLimit)) {
    outCurves.push_back(curve); // Fit is good!
    return;
  }

  // Fit is bad, split at the point of maximum error
  // Important: Prevent infinite recursion if split doesn't advance
  if (splitPoint == first || splitPoint == last) {
    // Fallback: simply bisect indices if geometric split fails
    splitPoint = (first + last) / 2;
  }

  fitRecursive(points, first, splitPoint, errorLimit, outCurves);
  fitRecursive(points, splitPoint, last, errorLimit, outCurves);

```

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
...
+    fitRecursive(chains[i], 0, chains[i].size() - 1, tolerance, result);
+
+    /*std::cout << "Curve: " << std::endl;
+    for (auto &c: result) {
+        std::cout << "{ (" << c.p0.x << ", " << c.p0.y << ") " << ", (" <<
+    c.p1.x << ", " << c.p1.y << ") " << ", (" << c.p2.x << ", " << c.p2.y <<
+    ")}" << ", " << std::endl;
+    }*/
+    results.push_back(result);
+  }
```
Ryan-Millard:
```suggestion
void fitCurveReduction(const std::vector<std::vector<Point>> &chains,
                       std::vector<std::vector<QuadBezier>> &results,
                       float tolerance) {
  for (const auto& chain : chains) {
    // Start recursion on the whole chain
    std::vector<QuadBezier> result;
    fitRecursive(chain, 0, chain.size() - 1, tolerance, result);

    results.push_back(result);
  }
```

src/wasm/modules/image/src/contours.cpp
Ryan-Millard: Do you need to round off all of these numbers? I sometimes just let it narrow because it isn't worth the bother of saving the fraction.

src/wasm/modules/image/src/contours.cpp
```diff
@@ -285,4 +289,637 @@ ContoursResult find_contours(const std::vector<uint8_t> &binary, int width,
...
+    int index;
+    Point newPos;
+  };
+  std::vector<Update> updatesA;
+  std::vector<Update> updatesB;
+
+  // --- Process Vector A (Snap to B's Geometry) ---
+  for (size_t i = 0; i < vecA.size(); ++i) {
+    int ax = (int)std::round(vecA[i].x);
+    int ay = (int)std::round(vecA[i].y);
```
Ryan-Millard:
```suggestion
    int ax{static_cast<int>(std::round(vecA[i].x))};
    int ay{static_cast<int>(std::round(vecA[i].y))};
```

src/wasm/modules/image/src/contours.cpp
```diff
@@ -285,4 +289,637 @@ ContoursResult find_contours(const std::vector<uint8_t> &binary, int width,
...
+        continue;
+
+      Point myPos = contours[c][p];
+      Point prev = contours[c][p - 1];
+      Point next = contours[c][p + 1];
+
+      // 1. Calculate My Laplacian Target (Where I want to go to be smooth)
+      Point myTarget;
+      myTarget.x = 0.25f * prev.x + 0.5f * myPos.x + 0.25f * next.x;
+      myTarget.y = 0.25f * prev.y + 0.5f * myPos.y + 0.25f * next.y;
```
Ryan-Millard: You frequently calculate this, so I think it would be a good idea to make a function for it.

src/wasm/modules/image/src/contours.cpp
```diff
@@ -285,4 +289,637 @@ ContoursResult find_contours(const std::vector<uint8_t> &binary, int width,
...
+      if (partnerCount > 0) {
+        float avgTx =
+            (myTarget.x + sumPartnerTargets.x) / (1.0f + partnerCount);
+        float avgTy =
+            (myTarget.y + sumPartnerTargets.y) / (1.0f + partnerCount);
+        targetPos[c][p] = {avgTx, avgTy};
+      } else {
+        // No partners, just smooth myself
+        targetPos[c][p] = myTarget;
+      }
```
Ryan-Millard: Excellent choice of diction.😂

src/wasm/modules/image/src/contours.cpp
```diff
@@ -285,4 +289,637 @@ ContoursResult find_contours(const std::vector<uint8_t> &binary, int width,
...
+    }
+
+    contours = nextContours;
+
+    // Smooth (respecting locks)
+    // for (int c = 0; c < (int)contours.size(); ++c) {
+    //     selectiveSmooth(contours[c], lockedMasks[c]);
+    //}
+    // coupledSmooth(contours, lockedMasks);
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/contours.cpp
```diff
@@ -285,4 +289,637 @@ ContoursResult find_contours(const std::vector<uint8_t> &binary, int width,
...
+  }
+
+  std::cout << "Packed with locked boundaries.\n";
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/contours.cpp
```diff
@@ -1,5 +1,7 @@
 
 #include "contours.h"
+#include <cmath>
+#include <iostream>
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/graph.cpp
```diff
@@ -88,8 +105,9 @@ void Graph::clear_unconnected_nodes() {
 
 void Graph::discover_edges(const std::vector<int32_t> &region_labels,
                            const int32_t width, const int32_t height) {
-  // Moore 4-connected neighbourhood
-  constexpr int8_t dirs[4][2]{{1, 0}, {-1, 0}, {0, 1}, {0, -1}};
+  // Moore 8-connected neighbourhood
+  constexpr int8_t dirs[8][2]{{1, 0}, {-1, 0},  {0, 1},  {0, -1},
+                              {1, 1}, {-1, -1}, {-1, 1}, {1, -1}};
```
Ryan-Millard: Why 8-connected neighbors now? Was 4 too inaccurate?

src/wasm/modules/image/src/graph.cpp
```diff
@@ -118,6 +136,206 @@ void Graph::discover_edges(const std::vector<int32_t> &region_labels,
...
+  
+  std::set<std::pair<int, int>> adjusted_neighbors{};
+
+  constexpr int8_t dirs[8][2]{{1, 0}, {-1, 0},  {0, 1},  {0, -1},
+                              {1, 1}, {-1, -1}, {-1, 1}, {1, -1}};
+
+  constexpr int8_t dirs5[16][2]{{0, -2}, {1, -2},  {2, -2},  {2, -1},
+                                {2, 0},  {2, 1},   {2, 2},   {1, 2},
+                                {0, 2},  {-1, 2},  {-2, 2},  {-2, 1},
+                                {-2, 0}, {-2, -1}, {-2, -2}, {-1, -2}};
```
Ryan-Millard: `dirs` was already defined beforehand and you didn't use `dirs5`.


```suggestion
```

src/wasm/modules/image/src/graph.cpp
```diff
@@ -118,6 +136,206 @@ void Graph::discover_edges(const std::vector<int32_t> &region_labels,
...
+        uint8_t val = neighborhood[y * bounds[2] + x];
+        std::cout << std::to_string(val);
+      }
+      std::cout<<std::endl;
+    }*/
+
+    // 0 = background, 1 = this node, 2+ = neighboring nodes
+    
+    // find touching edges
+    // std::vector<bool> addressed_pixels(neighborhood.size, false);
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/graph.cpp
```diff
@@ -128,9 +346,20 @@ void Graph::merge_small_area_nodes(const int32_t min_area) {
         std::copy(n->edges().begin(), n->edges().end(),
                   std::back_inserter(neighbors));
 
+        ImageLib::RGBPixel<uint8_t> col = n->color();
         // Sort by size -> a.area < b.area
+        // std::sort(neighbors.begin(), neighbors.end(),
+        //          [](Node_ptr a, Node_ptr b) { return a->area() < b->area();
+        //          });
+
+        // sort by size and color similarity
```
Ryan-Millard:
```suggestion
        // sort by size and color similarity
```

src/wasm/modules/image/src/kmeans_graph.cpp
```diff
@@ -140,46 +128,76 @@ void visualize_contours(const std::vector<std::vector<Point>> &contours,
...
+
+std::string contoursResultToSVG(const ColoredContours &result, const int width,
+                                const int height) {
+  std::ostringstream svg;
+  svg << "<svg xmlns=\"http://www.w3.org/2000/svg\" fill-rule=\"evenodd\" "
+         "width=\""
+      << width << "\" height=\"" << height << "\">\n";
+
+  // for (size_t i = 0; i < result.contours.size(); ++i) {
+  // std::string pathData = contourToSVGPath(result.contours[i]);
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/morphology.cpp
```diff
@@ -0,0 +1,738 @@
...
+  return img[y * w + x];
+}
+
+// Sets a pixel (no bounds check needed if logic is correct, but added for
+// safety)
+inline void setPixel(std::vector<uint8_t> &img, int w, int x, int y,
+                     uint8_t val) {
+  img[y * w + x] = val;
+}
+
```
Ryan-Millard:
```suggestion
inline void setPixel(std::vector<uint8_t> &img, int w, int x, int y,
                     uint8_t val) {
  if (x < 0 || x >= w || y < 0 || y >= h)
    return 0; // Boundary check
  img[y * w + x] = val;
}

```

src/wasm/modules/image/src/morphology.cpp
```diff
@@ -0,0 +1,738 @@
...
+        map[y * w + x] = 1; // Line (Body or Corner)
+      }
+    }
+  }
+  return map;
+}
+
+// Gaussian Smoothing (Kernel size 5, Sigma 1.0)
+// Preserves curves better than simple averaging
+void smoothChainGaussian(std::vector<Point> &chain) {
```
Ryan-Millard: What is the typical length of one of these chains? Are they are large or small?

src/wasm/modules/image/src/morphology.cpp
```diff
@@ -0,0 +1,738 @@
...
+
+float perpendicularDistSq(Point p, Point a, Point b) {
+  float l2 = (a.x - b.x) * (a.x - b.x) + (a.y - b.y) * (a.y - b.y);
+  if (l2 == 0)
+    return (p.x - a.x) * (p.x - a.x) + (p.y - a.y) * (p.y - a.y);
+  float t = ((p.x - a.x) * (b.x - a.x) + (p.y - a.y) * (b.y - a.y)) / l2;
+  t = std::max(0.0f, std::min(1.0f, t));
+  Point proj = {a.x + t * (b.x - a.x), a.y + t * (b.y - a.y)};
+  return (p.x - proj.x) * (p.x - proj.x) + (p.y - proj.y) * (p.y - proj.y);
+}
```
Ryan-Millard: I think Point needs to become a full class based on how it's being used.

src/wasm/modules/image/src/morphology.cpp
```diff
@@ -0,0 +1,738 @@
...
+          char c = '.';
+          if (val == 1) c = '+';       // Line
+          else if (val == 2) c = 'E';  // Endpoint
+          else if (val == 3) c = 'J';  // Junction
+          std::cout << c;
+      }
+      std::cout << "\n";
+  }
+  std::cout << "---------------------------------------------\n\n";
+  */
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/morphology.cpp
```diff
@@ -0,0 +1,738 @@
...
+      for (int i = 0; i < 8; ++i) {
+        int nx = cx + dx[i], ny = cy + dy[i];
+        if (nx < 0 || nx >= w || ny < 0 || ny >= h)
+          continue;
+        int nIdx = ny * w + nx;
+        if (!getPixel(skel, w, h, nx, ny))
+          continue;
+
+        // if (isJunction(nIdx)) { bestDir = i; break; } // Hit Junction
+        // if (!visited[nIdx] && topology[nIdx] != 3) { bestDir = i; break; } //
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/morphology.cpp
```diff
@@ -0,0 +1,738 @@
...
+  std::vector<std::vector<Point>> simplifiedChains;
+  for (const auto &c : chains) {
+    std::vector<Point> simple;
+    // Epsilon 1.0 - 1.5 is usually good for pixel art skeletons
+    ramerDouglasPeucker(c, 1.5f, simple);
+    simplifiedChains.push_back(simple);
+  }
+
+  return chains;
+}
```
Ryan-Millard:
```suggestion
```

src/wasm/modules/image/src/morphology.cpp
```diff
@@ -0,0 +1,738 @@
...
+    }
+  }
+
+  for (int i = 0; i < regions.size(); ++i) {
+    float area = getSignedArea(regions[i]);
+    std::cout << "Region " << i << ": Area = " << area << "\n";
+  }
+
+  return regions;
+}
```
Ryan-Millard:
```suggestion
```

Ryan-Millard:
I'm sorry for taking so long to review this. It was quite large and I had to find time where I could actually sit down for a while and read through everything. Your work is great!

src/wasm/modules/image/src/bezier.cpp
Krasner: `points` is the full contour. `first` and `last` specify which range in the contour we are fitting

Krasner:
bunch of stuff in `morphology.cpp` is experimental. It works but not being used right now. Is it worth keeping? Might be useful in the future

src/wasm/modules/image/src/graph.cpp
```diff
@@ -1,10 +1,27 @@
...
+  ImageLib::RGBPixel<float> af{static_cast<float>(a.red),
+                               static_cast<float>(a.green),
+                               static_cast<float>(a.blue)};
+  ImageLib::RGBPixel<float> bf{static_cast<float>(b.red),
+                               static_cast<float>(b.green),
+                               static_cast<float>(b.blue)};
+  return std::sqrt((af.red - bf.red) * (af.red - bf.red) +
+                   (af.green - bf.green) * (af.green - bf.green) +
+                   (af.blue - bf.blue) * (af.blue - bf.blue));
+}
```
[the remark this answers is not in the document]
Krasner: Done

src/wasm/modules/image/src/bezier.cpp
Krasner: so this isn't a function that outputs the same thing all the time. it's actually doing a least-squares fit

src/wasm/modules/image/src/bezier.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  return {numX / den, numY / den};
+}
+
+// --- Recursive Fit Function ---
+void fitRecursive(const std::vector<Point> &points, int first, int last,
+                  float errorLimit, std::vector<QuadBezier> &outCurves) {
+
+  // Base Case: Not enough points, just connect them
+  if (last - first < 2) {
```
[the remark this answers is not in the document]
Krasner: see comment on the recursive nature of this function. `points` is the full contour. `first` and `last` indicate the piece of contour we're fitting on.

src/wasm/modules/image/include/solver.h
```diff
@@ -0,0 +1,31 @@
...
+#define SOLVER_H
+
+#include "contours.h"
+#include <cmath>
+#include <iomanip>
+#include <numeric>
+#include <stdexcept>
+#include <vector>
+
+struct Point;
```
Ryan-Millard: Please separate this into its own file. Point has evolved into a reusable class that should be in its own file because we use it in multiple places (via redundant redefinitions and strange include via includes).

src/wasm/modules/image/include/solver.h
```diff
@@ -0,0 +1,31 @@
...
+
+class SavitzkyGolay {
+private:
+  int window_size_;
+  int m_; // half window size
```
Ryan-Millard: Why store this? If `m_` = `window_size_` / 2, why not make a function for this or compute it in-place? It just seems like there's a lot of extra work that goes into managing this variable. If `window_size_` ever changes, won't `m_` have to change?

Additionally, why is it called that? It seem to signal what it represents and I can imagine myself getting confused by it after a long day of coding because my head is a bit cloudy.

src/wasm/modules/image/src/kmeans_graph.cpp
```diff
@@ -140,46 +127,76 @@ void visualize_contours(const std::vector<std::vector<Point>> &contours,
...
+
+  std::cout << "result.curves.size() " << result.curves.size() << std::endl;
```
Ryan-Millard:
```suggestion
```

Ryan-Millard (changes requested):
Please clean up your includes. 

Try to have as few includes in a header file as possible and try to rather move them into the source file. Please also remove the unused includes.

src/wasm/modules/image/include/SavitskyGolay.h
```diff
@@ -0,0 +1,31 @@
...
+#include <numeric>
+#include <stdexcept>
+#include <vector>
+
+struct Point;
+
+class SavitzkyGolay {
+private:
+  int window_size_;
+  int m_; // half window size
```
Krasner: yes `m_ == window_size_ / 2` always. just short hand since `m_` is used in a lot of places. So just pre-calculating it in the constructor

Krasner:
i do have updates for smoother SVGs but they should probably be in a new PR.... this one is already too big

Ryan-Millard:
You work is absolutely brilliant.

dreamy-african-savannah


image

Ryan-Millard:
> > Are you interested in learning how to write better Docusaurus documentation?
> 
> yeah, do u have an easy explanation

#214 could be a good one to get started on if you're interested in it. The Getting Started page should probably be excluded, though, because #250 is about to change everything that applies in that section.

A lot of people have mentioned that it is difficult to read through and know where to go. Some of the pages, like the usage page and the commit style page should just be entirely deleted because they're a waste of energy. For example, I usually squash PRs, which means that the person's commit messages don't matter at all.

Additionally, some people have said that the sidebar goes over the fold, which isn't great because it hides away more important documentation.

PR #254, opened 2026-02-06
Feat/svg builder3

Krasner:
Note: rebasing is a mess. but this should be up-to-date with main.

A few updates to get smoother contours possibly:
1. SavitzkyGolay new method `filter_wrap_with_constraints` tries to prevent points on the borders from moving. https://github.com/Krasner/Img2Num/blob/0c610818b03a25a51e7321f40b29ebc1bc3b2742/src/wasm/modules/image/src/contours.cpp#L781C32-L781C60
   SG filter has some ringing to it, so further apply binomial smoothing
   https://github.com/Krasner/Img2Num/blob/0c610818b03a25a51e7321f40b29ebc1bc3b2742/src/wasm/modules/image/src/contours.cpp#L782
2. Support of Cubic Bezier curves. They add a 2nd control point, so they can make more complex shapes. Potentially helping with loops. https://github.com/Krasner/Img2Num/blob/0c610818b03a25a51e7321f40b29ebc1bc3b2742/src/wasm/modules/image/include/contours.h#L22
3. Since contours are all loops update any iteration over contour points to wrap around. Some previous code had
`for (int i = 1; i < pts.size() - 1; i++)` where looking back would be `i-1` and looking ahead is `i+1`. now all those loops are `for (int i = 0; i < pts.size() ; i++)` with the correct logic to wrap around. This prevents some distortion on the end points.

## Summary by CodeRabbit

* **Documentation**
  * Expanded graph module docs with a new Diagram section, multiple illustrative slides, and richer explanatory text.

* **New Features**
  * Added cubic Bezier curve fitting and SVG export alongside existing quadratic support.
  * Enhanced contour processing: improved smoothing, broader neighbor/circular handling, corner detection, and constrained smoothing that respects locked points.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -8674,13 +8674,13 @@
       }
     },
     "node_modules/enhanced-resolve": {
-      "version": "5.18.3",
-      "resolved": "https://registry.npmjs.org/enhanced-resolve/-/enhanced-resolve-5.18.3.tgz",
-      "integrity": "sha512-d4lC8xfavMeBjzGr2vECC3fsGXziXZQyJxD868h2M/mBI3PwAuODxAkLkq5HYuvrPYcUtiLzsTo8U3PgX3Ocww==",
+      "version": "5.19.0",
+      "resolved": "https://registry.npmjs.org/enhanced-resolve/-/enhanced-resolve-5.19.0.tgz",
+      "integrity": "sha512-phv3E1Xl4tQOShqSte26C7Fl84EwUdZsyOuSSk9qtAGyyQs2s3jJzComh+Abf4g187lUUAvH+H26omrqia2aGg==",
       "license": "MIT",
       "dependencies": {
         "graceful-fs": "^4.2.4",
-        "tapable": "^2.2.0"
+        "tapable": "^2.3.0"
       },
       "engines": {
         "node": ">=10.13.0"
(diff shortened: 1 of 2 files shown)


Krasner:
rebase is so bad 😠 
anyways the only changes are in some c++ headers and cpp files. There isn't much.

This is also optional... if you think it's not worth the merge then i'm ok with that.

src/wasm/modules/image/include/contours.h
```diff
@@ -11,13 +11,21 @@
 #include <cstdlib>
 #include <stdexcept>
 #include <vector>
+#include <numeric>
```
Ryan-Millard: Does this need to be included in the header? Header includes are implicitly included by any file that includes it.

src/wasm/modules/image/include/contours.h
```diff
@@ -11,13 +11,21 @@
...
   Point p1{0, 0}; // Control
   Point p2{0, 0}; // End
 };
 
+struct CubicBezier {
+  Point p0{0, 0}; // Start
+  Point p1{0, 0}; // Control 1
+  Point p2{0, 0}; // Control 2
+  Point p3{0, 0}; // End
+};
```
Ryan-Millard: Is using multiple structs like this a scalable approach? Will we ever want to go higher than $x^3$?

Ryan-Millard:
This doesn't look like a good addition to our code because it makes straight edges rounder, which pushes use closer to lossier output. We need to try to match the image as precisely as possible without creating bad SVGs.

I also want to take a moment to thank you for your sustained interest in this project. It's really nice to have someone else working on it and your contributions are always wonderful.

PR #307, opened 2026-04-10
feat(python bindings): enable python bindings and example

Krasner:
Enable python bindings as part of the c/cpp compilation. Update Dockerfile to have necessary python components.

**IMPORTANT: Update docker container on dockerhub!**

As part of building the c/cpp apps python bindings will be built.

```

cmake -B build-c-cpp/ .
cmake --build build-c-cpp
cd example_apps/console-py
python3 main.py

```

Note this will generate the `img2num` python lib in `build-c-cpp/bindings/python/`

## What was changed & why

Fixes: #<issue-number> 

## Summary by CodeRabbit

* **New Features**
  * Python bindings for image processing (in-place filters, k-means clustering, and labels→SVG export)
  * Python example app demonstrating end-to-end image processing and SVG/Saved outputs

* **Bug Fixes**
  * Corrected LAB→RGB color channel assignment

* **Chores**
  * Dev Docker image now includes Python build/runtime tooling and common packages
  * Build now includes native Python bindings, enables position-independent code, and emits Python-related build diagnostics


bindings/python/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,91 @@
+#include <pybind11/pybind11.h>
+#include <pybind11/numpy.h>
+#include <cstdlib>
+#include <cimg2num.h>
+
+namespace py = pybind11;
```
Ryan-Millard: I understand that it is easier for us to use this, but I think it would be better for consumers if we didn't. The difference between writing `py` versus `pybind11` are minimal for us, yet they could be very impactful for downstream consumers. Please will you remove this line.

This is a Python target, but someone may decide to extend our Python target via a bootstrap.

bindings/py/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <cimg2num.h>
+
+namespace py = pybind11;
+
+PYBIND11_MODULE(img2num, m) {
+    m.doc() = "Python bindings for the img2num C library";
+
+    // -----------------------------------------------------------------------
+    // In-place Image Modification Functions
+    // -----------------------------------------------------------------------
```
Ryan-Millard, line 12: Python expects that arguments passed are not altered, so this needs to choose one of two options to avoid being a weird library:
1. Leave these bindings as-is and add a smart wrapper package (what the JS library does)
2. Create a copy here instead of in a package and then operate on that copy.

bindings/python/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,91 @@
...
+    }, py::arg("data"), py::arg("width"), py::arg("height"), py::arg("k"), py::arg("max_iter"), py::arg("color_space"),
+       "Run K-Means clustering. Returns a tuple: (quantized_image, labels_array)");
+
+
+    m.def("labels_to_svg", [](py::array_t<uint8_t, py::array::c_style> data, py::array_t<int32_t, py::array::c_style> labels, int width, int height, int min_area) {
+        char* svg_c_str = img2num_labels_to_svg(
+            static_cast<const uint8_t*>(data.request().ptr),
+            static_cast<const int32_t*>(labels.request().ptr),
+            width, height, min_area
+        );
```
Ryan-Millard: This is just slightly more readable - the compiler should optimise the differences away. It's always good to have readable code.

```suggestion
        uint8_t* data{static_cast<const uint8_t*>(data.request().ptr)};
        uint8_t* labels{static_cast<const int32_t*>(labels.request().ptr)};
        char* svg_c_str = img2num_labels_to_svg(data, labels, width, height, min_area);
```

bindings/python/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,91 @@
...
+        if (!svg_c_str) {
+            throw std::runtime_error("img2num_labels_to_svg returned a null pointer.");
+        }
+
+        // Convert to Python string
+        py::str svg_py_str(svg_c_str);
+
+        // NOTE: If your C library dynamically allocates this string using malloc/calloc, 
+        // you MUST free it here to prevent a memory leak. If it returns a static pointer, remove this line.
+        free(svg_c_str);
```
Ryan-Millard: As a result of this, I think we should delay this PR because we need a function for them to call that frees it. I don't like the idea of C allocations inside of C++, so I want to change that - that also means that the entire C setup will need to change.

I want to make a function like `img2num_free_svg` instead.

bindings/py/CMakeLists.txt
Ryan-Millard, line 1: Looks good. Please just address CodeRabbit's concerns.

core/include/internal/cielab_impl.h
Ryan-Millard, line 1: What did this just do? I don't know if I understand your intentions, so I'd just like a quick explanation.

core/CMakeLists.txt
```diff
@@ -19,6 +19,11 @@ set(EMBEDDED_SHADERS_OUT "${CMAKE_BINARY_DIR}/embedded_shaders.h")
 set_source_files_properties(${EMBEDDED_SHADERS_OUT} PROPERTIES GENERATED TRUE)
 
 find_package(Python3 COMPONENTS Interpreter REQUIRED)
+message("Python Version: ${Python3_VERSION}")
+message("Python Version: ${Python3_VERSION_MAJOR}.${Python3_VERSION_MINOR}")
+message("Python Library: ${Python3_LIBRARIES}")
+message("Python Include: ${Python3_INCLUDE_DIRS}")
+
```
Ryan-Millard: Do we need this here? `core` is supposed to be independent of Python and its bindings.

example-apps/console-py/main.py
```diff
@@ -0,0 +1,30 @@
...
+import img2num
+
+"""
+Note: Images sent to img2num functions must be RGBA
+"""
+OUTDIR = "outputs"
+os.makedirs(OUTDIR, exist_ok=True)
+
+img = cv2.imread("../../IMG_5554.jpg")
+img = cv2.cvtColor(img, cv2.COLOR_BGR2RGBA) # VERY IMPORTANT!!!
```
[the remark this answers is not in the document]
Ryan-Millard: Valid concern. It's easy to miss and this is a great catch.

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
       context: .
       dockerfile: Dockerfile.dev
     container_name: img2num-dev
-    image: ryanmillard/img2num-dev:latest
+    # image: ryanmillard/img2num-dev:latest
```
Ryan-Millard: Please undo this. We'll need this once it gets merged.

Ryan-Millard:
I'm building it right now.😓

We need a Docker god to help us here. These builds take far too long.

Ryan-Millard:
How can I message you privately? As the days go by, I keep feeling the need to have a better way to communicate because this is becoming a bit tough.

A full Discord server might be overkill for us right now, but I'm not sure. I think we need a way to keep others updated and also message privately.

Krasner:
> These builds take far too long.

Yes :( Dawn takes forever

Ryan-Millard:
Thanks for the information - I’ve deleted your message containing your email address just in case.

I did email you and you can access the DockerHub repo here: https://hub.docker.com/repository/docker/ryanmillard/img2num-dev/general

Krasner:
> Thanks for the information - I’ve deleted your message containing your email address just in case.
> 
> I did email you and you can access the DockerHub repo here: https://hub.docker.com/repository/docker/ryanmillard/img2num-dev/general

Thanks! Got your email. Will reply a bit later

Dockerfile.dev
```diff
@@ -56,6 +59,10 @@ RUN corepack enable \
 ENV PATH=/root/.local/share/pnpm:$PATH
 ENV PNPM_STORE_DIR=/pnpm/store
 
+# Install some pip packages
+RUN python3 -m pip install numpy --break-system-packages
+RUN python3 -m pip install opencv-python --break-system-packages
```
Ryan-Millard: Why not have these locally? There is no real need to have a global, possibly damaging package installations.

Why not use a `requirements.txt`?

example-apps/console-py/main.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
+import cv2
+import os
+
+import sys
+sys.path.append("../../build-c-cpp/bindings/python/")
```
Ryan-Millard: There is no guarantee that the folder will be called `build-c-cpp`. For example, I name mine `build`.

Krasner:
@Ryan-Millard I updated the docker image (on dockerhub). seems to be working fine. It now included the necessary additions for python binding builds

Ryan-Millard:
> @Ryan-Millard I updated the docker image (on dockerhub). seems to be working fine. It now included the necessary additions for python binding builds

Thanks! I realised that our `Dockerfile.dev` isn't set up very well, which is slowing us down a lot. I'm busy working on fixing that right now, so I'll only be able to review this once I'm done. 

Please write some documentation for this - we need our users to know how to use the library once they import it.

I also noticed that you're referencing the Pybind build files directly from the `build-c-cpp` folder - that isn't great because other people might not use that folder name (mine is `build`). Please rather have a look at what I did for the WASM builds. I copied the files into the corresponding package location and created a convenience wrapper package around it in `packages/js`. See `build-wasm`:

```sh
ryan@Ryans-PC:~/projects/Img2Num$ ls packages/js/
build-wasm  imageToUint8ClampedArray.js  imageToUint8ClampedArray.test.js  index.js  package.json  safeWasmWrappers.js  tsconfig.typedoc.json  wasmClient.js  wasmWorker.js
ryan@Ryans-PC:~/projects/Img2Num$
```

Krasner:
Yeah I wasn't sure how to place the generated python lib... maybe i'll place it in `packages/python`

Krasner:
OK i made it so that the python library builds to the system (docker) python path:
```
Linking CXX shared module /usr/local/lib/python3.13/dist-packages/img2num.so
```

That way in `main.py` I can direct just do `import img2num` without specifying any path

Ryan-Millard:
I don't think it should build as a `.so` or be specific to the Docker image because downstream users won't use that. 

I'll clean it up - don't worry.

Ryan-Millard:
@Krasner, it's #308 if you're interested. I'm going to make it automatically push an image to Docker Hub every time a PR affects a Dockerfile and it'll allow us to just pull that image instead of building it ourselves. It'll be a major lifesaver later on.

Ryan-Millard:
@Krasner, I'm just going to wait until #308 is complete because the Docker build will be much faster after it is merged. It'll also be better for traceability because the images will be pushed to Docker Hub.

Krasner:
> @Krasner, I'm just going to wait until #308 is complete because the Docker build will be much faster after it is merged. It'll also be better for traceability because the images will be pushed to Docker Hub.

Yep no problem. I think you've got a good solution.

bindings/python/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    m.def("black_threshold_image", [](py::array_t<uint8_t, py::array::c_style> image, int width, int height, int num_thresholds) {
+        img2num_black_threshold_image(image.mutable_data(), width, height, num_thresholds);
+    }, py::arg("image"), py::arg("width"), py::arg("height"), py::arg("num_thresholds"),
+       "Apply black thresholding to the image (modifies image in-place)");
+
+    m.def("bilateral_filter", [](py::array_t<uint8_t, py::array::c_style> image, size_t width, size_t height, double sigma_spatial, double sigma_range, uint8_t color_space) {
+        img2num_bilateral_filter(image.mutable_data(), width, height, sigma_spatial, sigma_range, color_space);
+    }, py::arg("image"), py::arg("width"), py::arg("height"), py::arg("sigma_spatial"), py::arg("sigma_range"), py::arg("color_space"),
+       "Apply bilateral filter (modifies image in-place)");
```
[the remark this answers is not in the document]
Ryan-Millard: Shouldn't this problem rather be handled in the definition of this function inside of core?

bindings/python/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,91 @@
...
+            static_cast<const uint8_t*>(data_buf.ptr),
+            out_data_ptr,
+            out_labels_ptr,
+            width, height, k, max_iter, color_space
+        );
+
+        // Return a tuple of (out_data, out_labels)
+        return py::make_tuple(out_data, out_labels);
+    }, py::arg("data"), py::arg("width"), py::arg("height"), py::arg("k"), py::arg("max_iter"), py::arg("color_space"),
+       "Run K-Means clustering. Returns a tuple: (quantized_image, labels_array)");
```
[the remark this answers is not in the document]
Ryan-Millard: Shouldn't we rather just use default values?

Is this outdated?

Ryan-Millard:
@Krasner, I'm sorry for spending so much time on #308. I think it'll take quite a while because I don't want an annoying setup that barely works.

In the meantime, please would you have a look at the things we need to get done for [our milestones](https://github.com/Ryan-Millard/Img2Num/milestones). We're very close to v0.0.0 - we just need like 2 things.

If you're not interested in those, the Krita community seemed interested in a plugin, so that would be worth investigating. See [my post](https://krita-artists.org/t/raster-image-to-svg-image-conversion-vectorization-img2num/159735) on their site.

Inkscapes is also a possibility, but I think it would be best to test the waters with one first.

Krasner:
No worries. PR 308 is important.
Quickly glanced at the Krita thread - someone mentioned python so this PR would be good to merge in.

bindings/c/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ set(C_BINDINGS_SRC
 )
 
 add_library(CImg2Num ${C_BINDINGS_SRC})
+set_property(TARGET CImg2Num PROPERTY POSITION_INDEPENDENT_CODE ON)
```
Ryan-Millard: Isn't this better to have in `core`?

bindings/python/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,115 @@
...
+                throw std::runtime_error("img2num_labels_to_svg returned a null pointer.");
+            }
+
+            // Convert to Python string
+            py::str svg_py_str(svg_c_str);
+
+            // NOTE: If your C library dynamically allocates this string using malloc/calloc,
+            // you MUST free it here to prevent a memory leak. If it returns a static pointer,
+            // remove this line.
+            free(svg_c_str);
```
Ryan-Millard: Matches `std::malloc` in [`labels_to_svg.cpp` at line 251](https://github.com/Ryan-Millard/Img2Num/blob/b540cd83459112ce53e8bb1990f66bc4526fef66/core/src/internal/labels_to_svg.cpp#L251):

Image

```suggestion
            std::free(svg_c_str);
```

#266 was created to solve this problem, but C-style memory is okay for right now since we are using the C ABI. The only problem is that you can't `delete` it safely in C++.

bindings/python/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+file(GLOB_RECURSE BINDING_SRC CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp")
+
+find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
+
+find_package(PythonLibs REQUIRED)
+set(PYTHON_MODULE_EXTENSION ".so" CACHE INTERNAL "Cross python lib extension")
```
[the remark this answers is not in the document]
Ryan-Millard: This is the same as what is in [`core/CMakeLists.txt` file at line 21](https://github.com/Ryan-Millard/Img2Num/blob/b540cd83459112ce53e8bb1990f66bc4526fef66/core/CMakeLists.txt#L21) - it locates the system's Python package, correct?

bindings/python/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+    "${CMAKE_CURRENT_SOURCE_DIR}/include"
+)
+
+# include(GNUInstallDirs)
+# install(TARGETS img2num
+#     COMPONENT python
+#     LIBRARY DESTINATION "${Python3_SITEARCH}"
+#     RUNTIME DESTINATION "${Python3_SITEARCH}"
+# )
+
```
Ryan-Millard:
```suggestion
```

example-apps/console-py/main.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import cv2
+import os
+import sys
+
+import img2num
```
Ryan-Millard, line 4: How are you registering `img2num` as something importable? Are you installing it as a system header?

Ryan-Millard (commented):
Thank you @Krasner. This looks great!

core/CMakeLists.txt
Krasner: those were for debugging. will delete

Dockerfile.dev
Krasner: yes I can use requirements...

example-apps/console-py/main.py
Krasner: In CMakeLists.txt for bindings/python you will see this:
```
set_target_properties(img2num PROPERTIES
    LIBRARY_OUTPUT_DIRECTORY "${Python3_SITEARCH}"
)
```
This specifically places the compiled python binding (img2num.so) into the python3 dist-packages folder. Thus when the user does `import img2num` python will automatically look into it's own dist-packages folder and finds it.

core/include/internal/cielab_impl.h
Krasner: This is similar to the fix in `rgb_to_lab`  - the templated `lab_to_rgb` function takes arguments by value and not by reference. So `rgba.red`, `rgba.green`, `rgba.blue` don't get set directly by line 207. Variables r, g, b get assigned a value then they need to be written to the output `rgba` RGBAPixel

bindings/py/src/img2num_pybind.cpp
Krasner: OK changed to not modify in place

core/CMakeLists.txt
```diff
@@ -31,6 +36,8 @@ add_custom_target(Img2Num_shaders ALL
 
 add_library(Img2Num ${CORE_SRC})
 
+set_property(TARGET Img2Num PROPERTY POSITION_INDEPENDENT_CODE ON)
```
Krasner, line 35: There's one in `core/` too

core/CMakeLists.txt
Krasner: but Python is still needed for the embedded_shaders

bindings/python/src/img2num_pybind.cpp
Krasner: how would that be different than asking the user to use `std:free`?

`labels_to_svg` could return `std::string` which would manage it's own deallocation. in `wasm_wrapper.cpp` that can be converted to char* with `.c_str()`. i can try this out to see if it works and solves the `free` issue

Ryan-Millard: I'd like it to only use C++ memory inside `core` and what you're suggesting sounds great! I haven't had enough time to think on this, though.

> `labels_to_svg` could return `std::string` which would manage it's own deallocation. in `wasm_wrapper.cpp` that can be converted to char* with `.c_str()`. i can try this out to see if it works and solves the `free` issue

That sounds excellent, except for the WASM part - that should be handled in the C bindings instead. By handling it like that, we only need to change code in a single place. Emscripten expects C-usable code and, since we already have C bindings, we just re-export that C code to WASM.

The pipeline is this:

```mermaid
flowchart TD

subgraph Core Layer
    A["Core Library (C++)"]
end

subgraph C Bindings Layer
    B["C Bindings (bindings/c)"]
end

subgraph JS / WASM Layer
    C["JS Bindings (bindings/js)"]
end


A -->|Imported by| B
B -->|Imported by| C
```

Ryan-Millard: Additionally, it isn't very uncommon to have a special function for freeing memory created by a library in C - have a look: [line 112 in `example-apps/console-c/main.c`](https://github.com/Ryan-Millard/Img2Num/blob/b540cd83459112ce53e8bb1990f66bc4526fef66/example-apps/console-c/main.c#L112) and [STB's usage of their own library](https://github.com/nothings/stb/blob/28d546d5eb77d4585506a20480f4de2e706dff4c/tests/test_png_regress.c#L67).

core/include/internal/cielab_impl.h
Ryan-Millard: Okay, cool. I should've looked more carefully.

Dockerfile.dev
Ryan-Millard: Please do. It'll make package management far simpler for us - especially if someone adds a dependency and doesn't mention it to the community.
I wanted a package manager for the C and C++, but there are only ugly solutions for that. For the other things, I want to use package managers for simplicity - especially when people are new to the codebase.

Krasner:
@Ryan-Millard check the last commit. I changed `label_to_svg` to output `std::string` since it will handle it's own deallocation. python bindings now use `img2num.h` (i.e. c++ header) so no need for manual deallocation. `cimg2num.h` handles conversion from `std::string` to `char*`, and wasm_wrapper takes those results. JS side still needs a manual deallocation.

In our case as long as other use `img2num.h` as the main interface then `std::string` handles deallocation. `cimg2num.h` exists only for the wasm interface.

example-apps/console-py/main.py
Ryan-Millard: That sounds like a very strange way to do it. Is that the way that Pybind recommends?

I understand that this seems like a good idea, but I think it would be better to not go about it that way because it will make it tougher for us to offer proper support for the various languages. If you have a look at what I did for the JS package, you'll see that I built the WASM library and copied the build output to the `packages/js` folder so that the JS can import it. That is built a bit like a DLL, yet it is packaged with some JS code to enhance the experience for JS developers because it isn't nice to have to manage memory in a garbage-collected language.

Although most of that is not relevant in Python since Pybind handles that for us, it is unsafe for functions that return heap memory - like `labels_to_svg`. The reason I wrapped it entirely with JS is because I didn't want to change the entire `core` library just because JS requires a special setup - instead, I made a full wrapper in JS to accommodate the average JS developer.

---

My main concern is the distributability of the library. We will need to package is as a single unit.

Ryan-Millard:
> @Ryan-Millard check the last commit. I changed `label_to_svg` to output `std::string` since it will handle it's own deallocation. python bindings now use `img2num.h` (i.e. c++ header) so no need for manual deallocation. `cimg2num.h` handles conversion from `std::string` to `char*`, and wasm_wrapper takes those results. JS side still needs a manual deallocation.
> 
> In our case as long as other use `img2num.h` as the main interface then `std::string` handles deallocation. `cimg2num.h` exists only for the wasm interface.

@Krasner, that should be in a separate PR. Please rather: 
1. Create a new branch off of main
2. Cherry-pick the latest commit off of this branch (about `labels_to_svg`)
3. Place the cherry-picked commit in the new branch you just made
4. Submit a PR for it and reference #266 in the body of the PR

The reason this is better is because it allows us to have atomic commit histories when we look at the commits on the main branch (I typically squash as the merging strategy for PRs).

Krasner:
> > @Ryan-Millard check the last commit. I changed `label_to_svg` to output `std::string` since it will handle it's own deallocation. python bindings now use `img2num.h` (i.e. c++ header) so no need for manual deallocation. `cimg2num.h` handles conversion from `std::string` to `char*`, and wasm_wrapper takes those results. JS side still needs a manual deallocation.
> > In our case as long as other use `img2num.h` as the main interface then `std::string` handles deallocation. `cimg2num.h` exists only for the wasm interface.
> 
> @Krasner, that should be in a separate PR. Please rather:
> 
> 1. Create a new branch off of main
> 2. Cherry-pick the latest commit off of this branch (about `labels_to_svg`)
> 3. Place the cherry-picked commit in the new branch you just made
> 4. Submit a PR for it and reference [Refactor C++ core to use C++ idioms (new/delete) instead of malloc #266](https://github.com/Ryan-Millard/Img2Num/issues/266) in the body of the PR
> 
> The reason this is better is because it allows us to have atomic commit histories when we look at the commits on the main branch (I typically squash as the merging strategy for PRs).

Only it directly impacts the python bindings. These two topics can be updated together I think. PR title can change to incorporate the interface change

Krasner:
new graph looks like

```mermaid
flowchart TD

subgraph Core Layer
    A["Core Library (C++)"]
end

subgraph C Bindings Layer
    B["C Bindings (bindings/c)"]
end

subgraph JS / WASM Layer
    C["JS Bindings (bindings/js)"]
end

subgraph Python Layer
    D["Python Bindings (bindings/python)"]
end

A -->|Imported by| B
B -->|Imported by| C
A -->|Imported by| D
```

Ryan-Millard:
> new graph looks like
> 
> ```mermaid
> flowchart TD
> 
> subgraph Core Layer
>     A["Core Library (C++)"]
> end
> 
> subgraph C Bindings Layer
>     B["C Bindings (bindings/c)"]
> end
> 
> subgraph JS / WASM Layer
>     C["JS Bindings (bindings/js)"]
> end
> 
> subgraph Python Layer
>     D["Python Bindings (bindings/python)"]
> end
> 
> A -->|Imported by| B
> B -->|Imported by| C
> A -->|Imported by| D
> ```

@Krasner, I think there's been a miscommunication somewhere here.😅

I was talking about the implementation of C-style memory management from `bindings/js`. My argument was that it would be better to instead implement it in `bindings/c` since `bindings/js` depends on it and it would reduce future (and current) redundancy. What you showed me is not incorrect because it reflects the relevant updates from this PR, but it doesn't tackle my point about C-style memory since Pybind11 allows direct C++-to-Python bindings (unlike Emscripten).

Krasner:
> Pybind11 allows direct C++-to-Python bindings (unlike Emscripten)

Have you looked into embind and EMSCRIPTEN_BINDINGS
https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html

This would be good for the next PR

Ryan-Millard:
> Only it directly impacts the python bindings. These two topics can be updated together I think. PR title can change to incorporate the interface change

It does affect the Python example app, but is still unrelated to implementing Python bindings. The app is just a consumer of the library and the `labels_to_svg` API refactor will affect how the entire library is used (not just for Python).

For this PR, use `std::free`. Once it is merged, the PR that tackles #266 will overwrite that section and use the updates you created.

The reason for this is because the commit won't be atomic if we add it here. The ideal commits that should be yielded from the work you have done are:

### Commit 1 (Python support)
```txt
feat(python bindings): enable python bindings and example
```

### Commit 2 (`labels_to_svg` API refactor)
```txt
refactor(core)!: switch to C++-style memory in labels_to_svg

* Previously used malloc() and free() to handle dynamic char*
* Now uses new and delete for better safety
* Updated all related consumers (e.g., bindings, example apps) and documentation
* Does not affect C (CImg2Num) or JavaScript (img2num_wasm)
```

Note that the second one has an exclamation mark to signal that the API has been broken, meaning that consumers will need to update their usage of the library if they use the new version.

Krasner:
> > Only it directly impacts the python bindings. These two topics can be updated together I think. PR title can change to incorporate the interface change
> 
> It does affect the Python example app, but is still unrelated to implementing Python bindings. The app is just a consumer of the library and the `labels_to_svg` API refactor will affect how the entire library is used (not just for Python).
> 
> For this PR, use `std::free`. Once it is merged, the PR that tackles #266 will overwrite that section and use the updates you created.
> 
> The reason for this is because the commit won't be atomic if we add it here. The ideal commits that should be yielded from the work you have done are:
> 
> ### Commit 1 (Python support)
> ```
> feat(python bindings): enable python bindings and example
> ```
> 
> ### Commit 2 (`labels_to_svg` API refactor)
> ```
> refactor(core)!: switch to C++-style memory in labels_to_svg
> 
> * Previously used malloc() and free() to handle dynamic char*
> * Now uses new and delete for better safety
> * Updated all related consumers (e.g., bindings, example apps) and documentation
> * Does not affect C (CImg2Num) or JavaScript (img2num_wasm)
> ```
> 
> Note that the second one has an exclamation mark to signal that the API has been broken, meaning that consumers will need to update their usage of the library if they use the new version.

OK I un-did that last commit. Will put it in as a separate PR. Can you check that these python bindings are good to go?

Ryan-Millard:
> > Pybind11 allows direct C++-to-Python bindings (unlike Emscripten)
> 
> Have you looked into embind and EMSCRIPTEN_BINDINGS https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html
> 
> This would be good for the next PR

I did when I first set up the WASM compilation (that was during the first month of this project), but I chose not to for these reasons:
1. It's much slower than what we use right now.
2. Binding things like classes is strange, difficult, slow and completely unnecessary. When dealing with complex and weird-behaving code (the C ABI), it's better to keep the code simple and have a lot of manual work because you at least know what your code is doing - you don't really when binding a class because all of your C++ memory knowledge goes out the window because the runtime is in WASM but the JS garbage collector tries to get all friendly with it.😓
3. It's more difficult to maintain because it makes WASM exports prettier, which would encourage more of them and destabilize the API. This is defensive coding: if new developers don't have to dig through our documentation before finding how to make a WASM export, they'll be more likely to propose one without thinking of the consequences.
4. It breaks how JavaScript is supposed to be and it makes me feel sick when I look at it from a JavaScript lens. Take their [Memory Management documentation](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#memory-management) for example:
    ```js
    // Ignore the abhorrent use of `var` here because it's irrelevant
    // The `Module` object is strange, we typically just use `MyClass` in JS
    var x = new Module.MyClass;
    x.method();
    // Why on earth would we ever run `delete`? JavaScript has a Garbage Collector and this API is like dismantling RAII in C++ because it basically defeats the purpose of the language. Even if that wasn't the case, it is extremely strange because we always expect GC cleanups in JS, so it would be very easy to forget about it and have memory leaks.
    x.delete();
    ```

They did propose a RAII-like implementation (see [Automatic Memory Management](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#automatic-memory-management) and [ECMAScript Explicit Resource Management](https://tc39.es/proposal-explicit-resource-management/)), but - as with all lovely things in JS - there is little support for it (see [this conversation with ChatGPT](https://chatgpt.com/share/69de9353-c17c-83ea-80b8-354fab1a778d) and also the references it used at the end). Additionally, since RAII is not a concept in JavaScript, most people won't know how the hell to use it, thereby reducing the number of people using and contributing to the library - we're already in such a niche market.😭

Ultimately, I chose against using that stuff because the cons seemed greater than the pros. I did still take some inspiration from their ideas and made my own implementation of what they have there, which I think is way cleaner. That implementation is basically the entire `core` -> `packages/js` setup.

Ryan-Millard:
> OK I un-did that last commit. Will put it in as a separate PR. Can you check that these python bindings are good to go?

It's mostly good. Please just see these comments:
- https://github.com/Ryan-Millard/Img2Num/pull/307/changes#r3081573735
- https://github.com/Ryan-Millard/Img2Num/pull/307/changes#r3066823876 (click on "Committable Suggestion")
- https://github.com/Ryan-Millard/Img2Num/pull/307/changes#r3081274493

example-apps/console-py/requirements.txt
Krasner, line 1: @Ryan-Millard added here. Removed the pip installs from Dockerfile.dev

bindings/python/src/img2num_pybind.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    m.def(
+        "kmeans",
+        [](pybind11::array_t<uint8_t, pybind11::array::c_style> data, int32_t width, int32_t height, int32_t k,
+           int32_t max_iter, uint8_t color_space) {
+            pybind11::buffer_info data_buf = data.request();
+
+            // Allocate NumPy arrays for the outputs
+            auto out_data = pybind11::array_t<uint8_t>(data_buf.shape);
+            auto out_labels = pybind11::array_t<int32_t>({(ssize_t)height, (ssize_t)width});
```
Ryan-Millard: Please static_cast here.

```suggestion
            pybind11::array_t<uint8_t> out_data = static_cast<pybind11::array_t<uint8_t>>(data_buf.shape);
            pybind11::array_t<uint8_t> out_labels = static_cast<pybind11::array_t<int32_t>>({(ssize_t)height, (ssize_t)width});
```

Ryan-Millard:
I just noticed that you're importing the C library for this PR - why is that? Pybind11 supports C++, doesn't it?

Krasner:
> I just noticed that you're importing the C library for this PR - why is that? Pybind11 supports C++, doesn't it?

Yes, that was one of the changes I put into the last commit that I ended up reverting. The next PR that includes `std::string` output for `label_to_svg` will use the core Img2Num library

Ryan-Millard:
> > I just noticed that you're importing the C library for this PR - why is that? Pybind11 supports C++, doesn't it?
> 
> Yes, that was one of the changes I put into the last commit that I ended up reverting. The next PR that includes `std::string` output for `label_to_svg` will use the core Img2Num library

Why not use it in this PR? It shouldn't make a difference. If anything, things are slower now because `Img2Num` becomes a C library through `CImg2Num`, which adds an extra unnecessary layer on top.

I'm not talking about `std::malloc` and `std::free`. I am talking about the below in a C++ library:
```cpp
#include <cimg2num> // Why the C library and not the C++ one? We're in a C++ environment

void func() {
   img2num_kmeans(...);
}
```
```cmake
target_link_libraries(_img2num PRIVATE CImg2Num)
```

Above, we are explicitly using C wrappers when it would be faster and safer to use the C++ wrappers. The problem here is in the Python bindings and not about `labels_to_svg` at all.



We could just do:
```cpp
#include <img2num>

void func() {
   img2num::kmeans(...);
}
```
```cmake
target_link_libraries(_img2num PRIVATE Img2Num)
```

Krasner:
> > > I just noticed that you're importing the C library for this PR - why is that? Pybind11 supports C++, doesn't it?
> > 
> > 
> > Yes, that was one of the changes I put into the last commit that I ended up reverting. The next PR that includes `std::string` output for `label_to_svg` will use the core Img2Num library
> 
> Why not use it in this PR? It shouldn't make a difference. If anything, things are slower now because `Img2Num` becomes a C library through `CImg2Num`, which adds an extra unnecessary layer on top.
> 
> I'm not talking about `std::malloc` and `std::free`. I am talking about the below in a C++ library:
> 
> ```c++
> #include <cimg2num> // Why the C library and not the C++ one? We're in a C++ environment
> 
> void func() {
>    img2num_kmeans(...);
> }
> ```
> 
> ```cmake
> target_link_libraries(_img2num PRIVATE CImg2Num)
> ```
> 
> Above, we are explicitly using C wrappers when it would be faster and safer to use the C++ wrappers. The problem here is in the Python bindings and not about `labels_to_svg` at all.
> 
> We could just do:
> 
> ```c++
> #include <img2num>
> 
> void func() {
>    img2num::kmeans(...);
> }
> ```
> 
> ```cmake
> target_link_libraries(_img2num PRIVATE Img2Num)
> ```

Yes I know, that's what i had it the last (reverted) commit. Anyways I put it back in again.

Krasner:
merge?

Ryan-Millard:
> merge?

Almost. Your work is basically done. I'm just setting up a better monorepo management system.

I sent a message about that here like 3 hours ago, but I don't see it, so I probably forgot to hit send.😅

The main problem is how slow Dawn's builds are with the current setup, so I'm trying to complete #308 so that this PR can be easier to work on. Managing new package managers and stuff is painful because of Dawn.

We needs docs, though. I mustn't forget about that.

Ryan-Millard:
@Krasner, I just pushed an update (after 3 hours of waiting for it to build😭).

To use it, you'll need to pull the latest tag (give me about 5 minutes to push it because I forgot to), then run the below:

```sh
uv sync
```

```sh
uv run python3
```

```py
import numpy as np
import img2num

img = np.array([
    10, 20,
    30, 40
], dtype=np.uint8)

out = img2num.invert_image(img, 2, 2)
print(out)
```

My terminal showed this:
```sh
root@5a0fa47104cd:/usr/src/app# uv sync
Resolved 3 packages in 1ms
Checked 2 packages in 0.18ms
root@5a0fa47104cd:/usr/src/app# uv run python3
Python 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
... import img2num
...
... img = np.array([
...     10, 20,
...     30, 40
... ], dtype=np.uint8)
...
... out = img2num.invert_image(img, 2, 2)
... print(out)
...
[245 235 225  40]
>>> print("Mission Accomplished!")
Mission Accomplished!
>>>
root@5a0fa47104cd:/usr/src/app#
exit
```

---

We are now using a very strange package manager (`uv`) that works well with our setup - a C++ monorepo that integrates Python bindings.

Ryan-Millard:
Please note that I am not quite done with this. I need to clean things up and I don't think the example app works (I haven't tested it, though).

Krasner:
> pushed an update

I see it now

Ryan-Millard:
> > pushed an update
> 
> i don't see your push

I was being daft.😅

I pushed it now.

Ryan-Millard:
I think you need to make a .venv, first. Let me just check again.

Ryan-Millard:
@Krasner, my apologies for the sloppy merge commit - it didn't compile and I just amended that.

`uv` makes the `.venv` automatically, so you only need to run the below:
```sh
uv sync
```

```sh
uv run python3
```

```py
import numpy as np
import img2num

img = np.array([
    10, 20,
    30, 40
], dtype=np.uint8)

out = img2num.invert_image(img, 2, 2)
print(out)
```

Krasner:
explain how to build?

```

cmake -DBUILD_PYTHON=ON -B build-c-cpp/ .

```

gives me the following error:
```

CMake Error at /usr/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
  Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found
  version "3.13.5")
Call Stack (most recent call first):
  /usr/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-3.31/Modules/FindPython/Support.cmake:3997 (find_package_handle_standard_args)
  /usr/share/cmake-3.31/Modules/FindPython3.cmake:602 (include)
  bindings/python/CMakeLists.txt:8 (find_package)
```

Krasner:
even with `uv sync` I get the same error:
```
CMake Error at /usr/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
        Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found
        version "3.13.5")
      Call Stack (most recent call first):
        /usr/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
        /usr/share/cmake-3.31/Modules/FindPython/Support.cmake:3997 (find_package_handle_standard_args)
        /usr/share/cmake-3.31/Modules/FindPython3.cmake:602 (include)
        bindings/python/CMakeLists.txt:8 (find_package)
```

Ryan-Millard:
> even with `uv sync` I get the same error:
> 
> ```
> CMake Error at /usr/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
>         Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found
>         version "3.13.5")
>       Call Stack (most recent call first):
>         /usr/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
>         /usr/share/cmake-3.31/Modules/FindPython/Support.cmake:3997 (find_package_handle_standard_args)
>         /usr/share/cmake-3.31/Modules/FindPython3.cmake:602 (include)
>         bindings/python/CMakeLists.txt:8 (find_package)
> ```

Is that a local error or are you running it in the Docker container?

Krasner:
> Is that a local error or are you running it in the Docker container?

I'm using the latest Docker image (it looks like you updated it on dockerhub)

Krasner:
> > Is that a local error or are you running it in the Docker container?
> 
> I'm using the latest Docker image (it looks like you updated it on dockerhub)

wiping any old docker images on my machine and pulling again... could be wrong latest image

Krasner:
> > > Is that a local error or are you running it in the Docker container?
> > 
> > 
> > I'm using the latest Docker image (it looks like you updated it on dockerhub)
> 
> wiping any old docker images on my machine and pulling again... could be wrong latest image

nope - same issue

Krasner:
also since you've made the changes, python bindings are no longer built as part of the c/c++ build so that's why the CI tests pass

Ryan-Millard:
> also since you've made the changes, python bindings are no longer built as part of the c/c++ build so that's why the CI tests pass

That's okay - the Python bindings should be a separate thing like how I have set it up because the new Python build system does a lot of work for us without us needing to do too much.

I'm going to check this out shortly.

Ryan-Millard:
Please run this command (if you haven't already) and confirm that the image was actually deleted:
```sh
docker images # or docker image ls
```

I don't know what the problem could be because I deleted my image and pulled it again without any problems:

```sh
ryan@Ryans-PC:~/projects/Img2Num$ docker images
                                                                                                                                                                        i Info →   U  In Use
IMAGE   ID             DISK USAGE   CONTENT SIZE   EXTRA
ryan@Ryans-PC:~/projects/Img2Num$ ./img2num sh
Starting dev container...
[+] up 9/111
[+] up 14/14nmillard/img2num-dev:latest [⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿] Pulling                                                                                                                           13.6s
 ✔ Image ryanmillard/img2num-dev:latest Pulled                                                                                                                                         60.0s
 ✔ Network img2num_default              Created                                                                                                                                        0.2s
 ✔ Volume img2num_pnpm-store            Created                                                                                                                                        0.0s
 ✔ Container img2num-dev                Created                                                                                                                                        1.2s
uv sync
root@feeb847e43d1:/usr/src/app# uv sync
Resolved 3 packages in 1ms
      Built img2num @ file:///usr/src/app
Prepared 1 package in 21.24s
Uninstalled 1 package in 15ms
░░░░░░░░░░░░░░░░░░░░ [0/1] Installing wheels...                                                                                                                                             warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
         If the cache and target directories are on different filesystems, hardlinking may not be supported.
         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
Installed 1 package in 34ms
 ~ img2num==0.0.0 (from file:///usr/src/app)
root@feeb847e43d1:/usr/src/app# uv run python3
Python 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print(5)
5
>>>
root@feeb847e43d1:/usr/src/app#
```

Ryan-Millard:
If you used `./img2num destroy`, it wouldn't have worked because I accidentally made it reference the wrong image. I will fix that soon.

Krasner:
> Please run this command (if you haven't already) and confirm that the image was actually deleted:
> 
> ```shell
> docker images # or docker image ls
> ```
> 
> I don't know what the problem could be because I deleted my image and pulled it again without any problems:
> 
> ```shell
> ryan@Ryans-PC:~/projects/Img2Num$ docker images
>                                                                                                                                                                         i Info →   U  In Use
> IMAGE   ID             DISK USAGE   CONTENT SIZE   EXTRA
> ryan@Ryans-PC:~/projects/Img2Num$ ./img2num sh
> Starting dev container...
> [+] up 9/111
> [+] up 14/14nmillard/img2num-dev:latest [⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿] Pulling                                                                                                                           13.6s
>  ✔ Image ryanmillard/img2num-dev:latest Pulled                                                                                                                                         60.0s
>  ✔ Network img2num_default              Created                                                                                                                                        0.2s
>  ✔ Volume img2num_pnpm-store            Created                                                                                                                                        0.0s
>  ✔ Container img2num-dev                Created                                                                                                                                        1.2s
> uv sync
> root@feeb847e43d1:/usr/src/app# uv sync
> Resolved 3 packages in 1ms
>       Built img2num @ file:///usr/src/app
> Prepared 1 package in 21.24s
> Uninstalled 1 package in 15ms
> ░░░░░░░░░░░░░░░░░░░░ [0/1] Installing wheels...                                                                                                                                             warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
>          If the cache and target directories are on different filesystems, hardlinking may not be supported.
>          If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
> Installed 1 package in 34ms
>  ~ img2num==0.0.0 (from file:///usr/src/app)
> root@feeb847e43d1:/usr/src/app# uv run python3
> Python 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> print(5)
> 5
> >>>
> root@feeb847e43d1:/usr/src/app#
> ```

OK this works. Tried on a different machine. But `example_apps/console-py/main.py` needs `opencv-python`. Please add that to the uv package manager

Ryan-Millard:
> > Please run this command (if you haven't already) and confirm that the image was actually deleted:
> > ```shell
> > docker images # or docker image ls
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don't know what the problem could be because I deleted my image and pulled it again without any problems:
> > ```shell
> > ryan@Ryans-PC:~/projects/Img2Num$ docker images
> >                                                                                                                                                                         i Info →   U  In Use
> > IMAGE   ID             DISK USAGE   CONTENT SIZE   EXTRA
> > ryan@Ryans-PC:~/projects/Img2Num$ ./img2num sh
> > Starting dev container...
> > [+] up 9/111
> > [+] up 14/14nmillard/img2num-dev:latest [⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿] Pulling                                                                                                                           13.6s
> >  ✔ Image ryanmillard/img2num-dev:latest Pulled                                                                                                                                         60.0s
> >  ✔ Network img2num_default              Created                                                                                                                                        0.2s
> >  ✔ Volume img2num_pnpm-store            Created                                                                                                                                        0.0s
> >  ✔ Container img2num-dev                Created                                                                                                                                        1.2s
> > uv sync
> > root@feeb847e43d1:/usr/src/app# uv sync
> > Resolved 3 packages in 1ms
> >       Built img2num @ file:///usr/src/app
> > Prepared 1 package in 21.24s
> > Uninstalled 1 package in 15ms
> > ░░░░░░░░░░░░░░░░░░░░ [0/1] Installing wheels...                                                                                                                                             warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
> >          If the cache and target directories are on different filesystems, hardlinking may not be supported.
> >          If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
> > Installed 1 package in 34ms
> >  ~ img2num==0.0.0 (from file:///usr/src/app)
> > root@feeb847e43d1:/usr/src/app# uv run python3
> > Python 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0] on linux
> > Type "help", "copyright", "credits" or "license" for more information.
> > >>> print(5)
> > 5
> > >>>
> > root@feeb847e43d1:/usr/src/app#
> > ```
> 
> OK this works. Tried on a different machine. But `example_apps/console-py/main.py` needs `opencv-python`. Please add that to the uv package manager

Okay. I'll do that now.

It's also strange that it didn't work on the first machine you used. Please will you investigate that and let me know if the problem is something I should fix.

Krasner:
@Ryan-Millard problem or question about uv...
when I change something in c++ and rebuild, that change it not reflected in the python package. specifically `uv sync` doesn't re-compile the python binding

Ryan-Millard:
> @Ryan-Millard problem or question about uv... when I change something in c++ and rebuild, that change it not reflected in the python package. specifically `uv sync` doesn't re-compile the python binding

Run this:
```sh
uv build --package img2num --wheel
```

Ryan-Millard:
TODO:
- Documentation (including docstrings)
- CI workflow
- numpy img.shape default support

Do we need anything else?

Krasner:
> TODO:
> 
> * Documentation
> * CI workflow
> 
> Do we need anything else?

Just those

packages/py/img2num/__init__.py
Krasner, line 1: @Ryan-Millard  why just gaussian_blur? test?

Ryan-Millard: I forgot to remove it. Does it print that line for you?

Ryan-Millard: I forgot to remove it. Does it print that line for you?

Krasner: no it doesn't print when I do `img2num.gaussian_blur_fft`

Krasner: it actually does.... had to pull latest commits
```
>>> img2num.gaussian_blur_fft(np.zeros((512,512,4),np.uint8), 512, 512, 50.0)
Running gaussian blur
array([[[0, 0, 0, 0],
        [0, 0, 0, 0],
        [0, 0, 0, 0], ...
```

Ryan-Millard: That's good. I'll remove it shortly. I just added it to verify that the Python wrapper in `packages/py` was actually being used.

Ryan-Millard:
I sorted out the print statement, but I can't get it to update and recognise that `packages/py/img2num/__init__.py` was edited without removing the `.venv` file. I've spent way too long on this and feel like I'm getting nowhere.

It needs to update whenever a change is made to a file in `packages/py`, yet it chooses not to.

Krasner:
`uv sync --reinstall` will just rebuild it all

Ryan-Millard:
> `uv sync --reinstall` will just rebuild it all

Does that also reflect updates to `__init__.py`?

Krasner:
> > `uv sync --reinstall` will just rebuild it all
> 
> Does that also reflect updates to `__init__.py`?

Yes, I tested it with removing and returning the `gaussian_blur_fft` thing you had.

Ryan-Millard:
> > > `uv sync --reinstall` will just rebuild it all
> > 
> > Does that also reflect updates to `__init__.py`?
> 
> Yes, I tested it with removing and returning the `gaussian_blur_fft` thing you had.

I've configured something poorly in `packages/py/pyproject.toml`. The Python shouldn't need a fancy rebuild each time it changes like the C++ does.

Krasner:
> > > > `uv sync --reinstall` will just rebuild it all
> > > 
> > > 
> > > Does that also reflect updates to `__init__.py`?
> > 
> > 
> > Yes, I tested it with removing and returning the `gaussian_blur_fft` thing you had.
> 
> I've configured something poorly in `packages/py/pyproject.toml`. The Python shouldn't need a fancy rebuild each time it changes like the C++ does.

Seems like a failing of uv and not something you're doing wrong.

Ryan-Millard:
> > > > > `uv sync --reinstall` will just rebuild it all
> > > > 
> > > > 
> > > > Does that also reflect updates to `__init__.py`?
> > > 
> > > 
> > > Yes, I tested it with removing and returning the `gaussian_blur_fft` thing you had.
> > 
> > I've configured something poorly in `packages/py/pyproject.toml`. The Python shouldn't need a fancy rebuild each time it changes like the C++ does.
> 
> Seems like a failing of uv and not something you're doing wrong.

Oh. I thought it was a problem with how I set up scikit builds in the package. Do you have any links for me to read about it?

Ryan-Millard:
@Krasner, please try my latest commit. It rebuilds everything when you run an app that uses the library - not the most optimal but I'm tired of looking for solutions to something that doesn't matter that much.

I also added a dev script to the `console-py` package, so you can run the below with it:
```sh
uv run --package console-py dev test.jpg
```

Let me know if you like it or if you preferred the `uv sync --reinstall --package packages/py` version more.

---

I started off wanting `bindings/py` to control its own build like `bindings/js` does, but I failed to get there. It might be worth having a look at in the future, but I am honestly so tired of all of this right now. #308 wasted my time so much and I just want to take a break from the hard stuff for like a week or so to avoid burning out. After this PR, I'll do some documentation because we need that and its relatively simple to work on.

Krasner:
@Ryan-Millard thanks for the large effort you put in with the uv packaging!

Krasner:
```
root@c3e90fa53949:/usr/src/app# uv sync
Resolved 12 packages in 1ms
Prepared 3 packages in 166ms
░░░░░░░░░░░░░░░░░░░░ [0/3] Installing wheels...                                                          warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
         If the cache and target directories are on different filesystems, hardlinking may not be supported.
         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
Installed 3 packages in 28ms
      Built img2num @ file:///usr/src/app/packages/py
Prepared 1 package without build isolation in 11.52s
Uninstalled 3 packages in 67ms
Installed 1 package in 23ms
 - console-py==0.0.0 (from file:///usr/src/app/example-apps/console-py)
 ~ img2num==0.0.0 (from file:///usr/src/app/packages/py)
 - opencv-python==4.13.0.92
 + packaging==26.1
 + pathspec==1.0.4
 + scikit-build-core==0.12.2
root@c3e90fa53949:/usr/src/app#
root@c3e90fa53949:/usr/src/app#
root@c3e90fa53949:/usr/src/app#
root@c3e90fa53949:/usr/src/app# uv run --package console-py dev test.jpg
░░░░░░░░░░░░░░░░░░░░ [0/2] Installing wheels...                                                          warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
         If the cache and target directories are on different filesystems, hardlinking may not be supported.
         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
Installed 2 packages in 232ms
Traceback (most recent call last):
  File "/usr/src/app/.venv/bin/dev", line 4, in <module>
    from main import main
ModuleNotFoundError: No module named 'main'
root@c3e90fa53949:/usr/src/app#
```

Ryan-Millard:
> ```
> root@c3e90fa53949:/usr/src/app# uv sync
> Resolved 12 packages in 1ms
> Prepared 3 packages in 166ms
> ░░░░░░░░░░░░░░░░░░░░ [0/3] Installing wheels...                                                          warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
>          If the cache and target directories are on different filesystems, hardlinking may not be supported.
>          If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
> Installed 3 packages in 28ms
>       Built img2num @ file:///usr/src/app/packages/py
> Prepared 1 package without build isolation in 11.52s
> Uninstalled 3 packages in 67ms
> Installed 1 package in 23ms
>  - console-py==0.0.0 (from file:///usr/src/app/example-apps/console-py)
>  ~ img2num==0.0.0 (from file:///usr/src/app/packages/py)
>  - opencv-python==4.13.0.92
>  + packaging==26.1
>  + pathspec==1.0.4
>  + scikit-build-core==0.12.2
> root@c3e90fa53949:/usr/src/app#
> root@c3e90fa53949:/usr/src/app#
> root@c3e90fa53949:/usr/src/app#
> root@c3e90fa53949:/usr/src/app# uv run --package console-py dev test.jpg
> ░░░░░░░░░░░░░░░░░░░░ [0/2] Installing wheels...                                                          warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
>          If the cache and target directories are on different filesystems, hardlinking may not be supported.
>          If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
> Installed 2 packages in 232ms
> Traceback (most recent call last):
>   File "/usr/src/app/.venv/bin/dev", line 4, in <module>
>     from main import main
> ModuleNotFoundError: No module named 'main'
> root@c3e90fa53949:/usr/src/app#
> ```

I'll have a look at this tomorrow. It's probably not a big deal.

Ryan-Millard:
Hi @Krasner, #308 had a few problems related to race conditions that I didn't know how to solve and forgot about them after only wanting to leave them briefly, but I solved them in #310.

The issue now is that I forgot that secrets aren't allowed to be used in PRs that come from forks, so I'm going to have to research a solution to that. I think the best one is to push to a GitHub package instead.

---

For now, please will you look at documentation automation via doc strings or similar for the Python so we don't need to work as hard to maintain that stuff. Look at how I did it for C++ (doxygen) and also JS (docusaurus plugin). I'm not expecting you.

Ryan-Millard:
@Krasner, I'm sorry about this - the Docker setup kind of broke things and I wound up wasting time with the assumption that things would work since they worked for my pull requests. That was an oversight because workflows have varying contexts and only mine run in this repository's context (meaning it has more permissions). I'm looking at a solution to the Docker scenario using artifacts, but it is taking a while.

Krasner:
No worries. Does the github packages approach seem to be working better?

Ryan-Millard:
> No worries. Does the github packages approach seem to be working better?

No. I'm in the same boat - I set all of it up for nothing.😓

The only benefit of the package is that it makes the project look better at a glance because we will be frequently pushing to it (whenever `main` is updated or we release a tag).

---

The best solution to the Docker problem is to use artefacts in the form of `.tar.gz` files that we download from GitHub and unzip - it's not very nice, but it'll save us from having to build every image ourselves.

Krasner:
take a look at this: https://michaelheap.com/access-secrets-from-forks/

Krasner:
@Ryan-Millard this need a Build Python CI test otherwise we won't catch an error if the docker image doesn't contain the pybind additions

Ryan-Millard:
My apolocheese. I messed up in #312  and forgot to add an environment variable.

Ryan-Millard:
> @Ryan-Millard this need a Build Python CI test otherwise we won't catch an error if the docker image doesn't contain the pybind additions

Yes. We need to add that. I still have a checklist for what this PR needs. It's hidden somewhere up there.

Ryan-Millard:
HOOOORAAAAYYY!!!!!🥳🥳

https://github.com/Ryan-Millard/Img2Num/actions/runs/24753630859

Krasner:
> HOOOORAAAAYYY!!!!!🥳🥳
> 
> https://github.com/Ryan-Millard/Img2Num/actions/runs/24753630859

but even though it is rebuilding the docker image, the other CI tests continued without waiting... and used the last available image

Krasner:
@Ryan-Millard in docker-compose the build block is lost:
```
build:
      context: .
      dockerfile: Dockerfile.dev
    container_name: img2num-dev
```
is this a mistake?

Krasner:
> ## 🐳 Docker image built successfully!
> ### Image
> [`ghcr.io/ryan-millard/img2num-dev:pr-307`](https://github.com/ryan-millard/Img2Num/pkgs/container/img2num-dev/pr-307)
> 
> ### Run it locally:
> ```shell
> IMG2NUM_IMAGE=ghcr.io/ryan-millard/img2num-dev:pr-307 ./img2num sh
> ```

@coderabbitai this actually is untagged ghcr.io/ryan-millard/img2num-dev@sha256:403ea1f2061b63757556cae0088d4694f737bf67ce3eb999d4dec4b3152010cd

why was it not tagged as pr-307?

Krasner:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> Let me look at the CI workflow files to understand how the Docker image tagging is handled.
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> # Find workflow files related to Docker
> fd -e yml -e yaml . .github/workflows/ --exec echo {}
> 
> # Look for docker-related workflow content
> rg -l "docker" .github/workflows/ 2>/dev/null || echo "No .github/workflows dir or no matches"
> ```
> 
> Length of output: 531
> 
> 🏁 Script executed:
> 
> ```
> #!/bin/bash
> cat .github/workflows/docker-ci.yml
> echo "---COMMENTER---"
> cat .github/workflows/commenter.yml
> ```
> 
> Length of output: 12547
> 
> `@Krasner`, I found the root cause. In `.github/workflows/docker-ci.yml`, the `Build & Push Docker image` step has `push: false`:
> 
> ```yaml
> - name: Build & Push Docker image
>   uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f
>   with:
>     context: .
>     file: Dockerfile.dev
>     push: false          # ← this is the bug
>     tags: ${{ steps.meta.outputs.tags }}
>     cache-from: ...
>     cache-to: type=registry,ref=${{ steps.meta.outputs.cache_ref }},mode=max
> ```
> 
> Because `push: false`, the runner builds the image locally and **only the BuildKit cache** is pushed to the registry (via `cache-to`). The BuildKit cache blob has no tag — it's what you're seeing as the bare SHA256 digest. The named tag `pr-307` is never actually pushed to GHCR, making the bot's success comment misleading.
> 
> The fix is straightforward — change `push: false` to `push: true`:
> 
> ```yaml
> - name: Build & Push Docker image
>   uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f
>   with:
>     context: .
>     file: Dockerfile.dev
>     push: true
>     tags: ${{ steps.meta.outputs.tags }}
>     cache-from: |
>       type=registry,ref=${{ steps.meta.outputs.cache_ref }}
>       type=registry,ref=${{ env.GHCR_IMAGE }}:buildcache
>     cache-to: type=registry,ref=${{ steps.meta.outputs.cache_ref }},mode=max
> ```
> 
> This will push both the proper tagged image (e.g., `ghcr.io/ryan-millard/img2num-dev:pr-307`) **and** the cache layer, so the bot comment will accurately reflect what's available in GHCR.
> 
> 🧠 Learnings used

@coderabbitai now I see cache-pr-307 tag but not pr-307

Krasner:
@Ryan-Millard the main issue is that set docker image script isn't finding the correct GHCR image. Maybe it's some permission issue, but the `GHCR_STATUS` checks always return 404, so then the script defaults to using the `:main` tag, which is not the correct image for this PR. 
Can you either verify that the 404 is coming from some permissions issue, or that the curl command is incorrect. In the img2num ghcr list I do see `cache-pr-307` present

Ryan-Millard:
> @Ryan-Millard in docker-compose the build block is lost:
> 
> ```
> build:
>       context: .
>       dockerfile: Dockerfile.dev
>     container_name: img2num-dev
> ```
> 
> is this a mistake?

I removed it because I want docker compose to favour pulling images rather than building them locally.

> @coderabbitai this actually is untagged ghcr.io/ryan-millard/img2num-dev@sha256:403ea1f2061b63757556cae0088d4694f737bf67ce3eb999d4dec4b3152010cd
>
> why was it not tagged as pr-307?

Because a tag is like a pointer - it doesn't directly hold the information. GHCR exposes this (unlike Docker Hub). 

---

I did have a mistake - I forgot to set `push: true` in the workflow. I'm having a look at the problems now.

Ryan-Millard:
@Krasner, may I reset the `.github/workflows` folder since the problem is no longer present?

- [Workflow run](https://github.com/Ryan-Millard/Img2Num/actions/runs/24793113726)
- GHCR:
    - [`pr-307`](https://github.com/users/Ryan-Millard/packages/container/img2num-dev/814889610?tag=pr-307)
    - [`cache-pr-307`](https://github.com/users/Ryan-Millard/packages/container/img2num-dev/814889540?tag=cache-pr-307)
- Docker Hub:
    - Nothing because GHCR is used here.


---

I'll run this to reset it:

```bash
git checkout origin main -- .github/
```

Krasner:
> @Krasner, may I reset the `.github/workflows` folder since the problem is no longer present?

yes

Krasner:
> > @Krasner, may I reset the `.github/workflows` folder since the problem is no longer present?
> 
> yes

just keep the Python ci test

Krasner:
still not pulling `pr-307` ghcr image
image

Ryan-Millard:
> still not pulling `pr-307` ghcr image image

CodeRabbit mentioned this in #320. I'll make that fix now.

Ryan-Millard:
Sorry, #320 doesn't cover that, but it's the same thing.

Krasner:
@Ryan-Millard python ci build works now!

Krasner:
Let's merge this?

Ryan-Millard:
> Let's merge this?

Sure. We can handle the below in another PR.

@coderabbitai, plase make issues for the below (they are all Python related for PR #307):

**TODO:**
- Documentation (including docstrings)
- numpy img.shape default support
- `.coderabbit.yaml` updates

Also, do you think we need anything else?

Ryan-Millard:
@coderabbitai

Please suggest a conventional commit for this PR.

PR #314, opened 2026-04-19
fix(destroy): persist resolved image name in state file

### LINKED ISSUE #313: fix: `./img2num destroy` command does not track and remove the correct Docker image
> [!TIP]
> This issue hasn't been claimed yet. Comment `/take` if you'd like to work on it!

---

## Summary

The `destroy` command in both `img2num` (bash) and `img2num.ps1` (PowerShell) is supposed to remove the Docker image after tearing down containers, volumes, and orphans. It does not reliably do so.

## Problem

Both scripts attempt to remove either a hardcoded image name or `$IMG2NUM_IMAGE` / `$env:IMG2NUM_IMAGE` at the moment `destroy` is invoked. This fails in the following scenario:

```
IMG2NUM_IMAGE=ghcr.io/ryan-millard/img2num-dev:pr-312 ./img2num sh
# ... do some work, then later ...
./img2num destroy
```

At the time `destroy` is called, `IMG2NUM_IMAGE` is no longer set in the shell environment. The scripts therefore either:
- Fall back to the hardcoded default (`ryanmillard/img2num-dev:latest`), removing the **wrong** image, or
- Remove nothing at all

There is no mechanism to track which image was last used.

## Expected behaviour

`destroy` should always remove whichever image was actually pulled and used, regardless of whether `IMG2NUM_IMAGE` is set at the time of the `destroy` call.

## Root cause

Neither script persists the resolved image name between invocations. The image is only knowable at the time the container is started — not when it is destroyed.

## Requirements for a fix

- Must work on **bash** (Linux/macOS) and **PowerShell** (Windows via `img2num.ps1`)
- Must handle `IMG2NUM_IMAGE` being set at start time but absent at destroy time
- Must not require Docker daemon introspection if avoidable (i.e. prefer not relying on `docker inspect`)
- Should be simple — a lock file or state file written at `up` time and read at `destroy` time is the likely approach

## Proposed approach (for discussion)

Write the resolved image name to a local state file (e.g. `.img2num-state`) when the container is started, and read from it during `destroy`. Both scripts should agree on the same file path and format.

```
# Written by sh/pnpm/any up command:
IMAGE=ghcr.io/ryan-millard/img2num-dev:pr-312
```

`destroy` then reads this file, removes the image, and deletes the state file.

This file should be added to `.gitignore`.

tmchow:
## Summary

Closes #313. `./img2num destroy` was removing either a hardcoded image name or the currently-set `$IMG2NUM_IMAGE`, which misbehaves in the exact workflow the issue describes -- `destroy` run in a fresh shell, long after `IMG2NUM_IMAGE` was exported for the `up` call. This PR persists the resolved image name to a local state file at start time and reads it back at destroy time.

## Reproduction (from #313)

```
IMG2NUM_IMAGE=ghcr.io/ryan-millard/img2num-dev:pr-312 ./img2num sh
# ... do some work, then later in a fresh shell ...
./img2num destroy
# -> Before this PR: removes 'ryanmillard/img2num-dev:latest' (the hardcoded default) or nothing at all
# -> After this PR:  removes 'ghcr.io/ryan-millard/img2num-dev:pr-312' via .img2num-state
```

The PowerShell entry-point (`img2num.ps1`) had the identical bug with its own hardcoded fallback (`img2num-dev:latest`). Both scripts are fixed here.

## Approach

Follows the "proposed approach" in the issue body:

1. **Resolve the image name once per invocation**, mirroring `docker-compose.yml` (`${IMG2NUM_IMAGE:-ryanmillard/img2num-dev:latest}`). Bash uses `IMG2NUM_RESOLVED_IMAGE`; PowerShell uses `$IMG2NUM_RESOLVED_IMAGE`.
2. **Write `IMAGE=<resolved>
` to `./.img2num-state`** whenever the container is brought up (the `run_in_container` / `Run-InContainer` wrappers), so subsequent `destroy` calls in a different shell can reconstruct what was actually running.
3. **`destroy` reads the state file**, `docker rmi`s that image, and deletes the state file. If the state file is absent (pre-state-file installs, or the user ran `destroy` without ever running `up`), it falls back to the current-shell resolved value -- same behaviour as before, so no regression for existing users.
4. **`.gitignore`** learns about `.img2num-state`.

No `docker inspect` dependency, per the issue's "prefer not relying on docker inspect" requirement.

## Verification

Ran the bash prologue functions against the three scenarios that define the fix:

```
$ bash /tmp/test-img2num.sh
Scenario 1 (env+save): resolved=ghcr.io/ryan/custom:v1, state contents:
IMAGE=ghcr.io/ryan/custom:v1
Scenario 2 (env gone, load): fallback=ryanmillard/img2num-dev:latest, loaded=ghcr.io/ryan/custom:v1
Scenario 3 (no state): loaded=ryanmillard/img2num-dev:latest
```

- Scenario 1: the custom `IMG2NUM_IMAGE` persists to disk (not the hardcoded default).
- Scenario 2: after losing the env var, `load_state_image` still returns the custom image.
- Scenario 3: with the state file absent, behaviour falls back to the env-or-default resolution -- i.e. strictly no worse than current.

`bash -n img2num` is clean (`pwsh` isn't installed locally so the `img2num.ps1` change couldn't be linted here, but it mirrors the bash version line-for-line -- same `Save-State` / `Load-StateImage` / `Run-InContainer.Save-State()` / destroy-reads-then-removes flow).

## Scope note

Out of scope for this PR: the pre-existing mismatch between the bash default (`ryanmillard/img2num-dev:latest`) and the PowerShell hardcoded fallback inside the old `destroy` branch (`img2num-dev:latest`). Both scripts now resolve via the docker-compose default so the mismatch is incidentally fixed for the destroy path, but the PS1 script didn't otherwise reference `IMG2NUM_IMAGE` anywhere else -- that broader parity pass is a separate issue worth, not something I expanded scope to cover here.

Closes #313

---

This contribution was developed with AI assistance (Claude Code).

## Summary by CodeRabbit

* **Improvements**
  * Docker container startup now persists the resolved image name, enabling accurate identification and removal of the correct image during cleanup operations instead of a default fallback.

* **Chores**
  * Updated version control configuration to ignore temporary state files generated during container operations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ logs
 *:Zone.Identifier
 
 .env
+
+# img2num destroy state
+.img2num-state
--- a/img2num
+++ b/img2num
@@ -2,8 +2,33 @@
 MODE="$1"
 shift || true
 
+# Resolved image name for this invocation. Mirrors docker-compose.yml default
+# so `IMG2NUM_IMAGE=... ./img2num sh` and a bare `./img2num sh` both resolve.
+IMG2NUM_RESOLVED_IMAGE="${IMG2NUM_IMAGE:-ryanmillard/img2num-dev:latest}"
+IMG2NUM_STATE_FILE="$(dirname "$0")/.img2num-state"
+
+# Record the image we're about to start so that a later `./img2num destroy`
+# in a fresh shell -- where IMG2NUM_IMAGE may no longer be set -- still knows
+# which image to remove. Keep it simple: a single KEY=value line.
+save_state() {
+  printf 'IMAGE=%s\n' "$IMG2NUM_RESOLVED_IMAGE" > "$IMG2NUM_STATE_FILE"
+}
+
+# Read the image name persisted by a previous `up`. Falls back to the current
+# resolved value (env var or hardcoded default) if the state file is missing
+# -- preserves pre-state-file behaviour.
+load_state_image() {
+  if [ -f "$IMG2NUM_STATE_FILE" ]; then
+    # shellcheck disable=SC1091
+    . "$IMG2NUM_STATE_FILE"
+    [ -n "${IMAGE:-}" ] && printf '%s\n' "$IMAGE" && return
+  fi
+  printf '%s\n' "$IMG2NUM_RESOLVED_IMAGE"
+}
+
 # Run a command inside the dev container
 run_in_container() {
+  save_state
   docker compose up -d dev
   docker compose exec dev "$@"
 }
@@ -26,7 +51,11 @@ case "$MODE" in
   down) docker compose down ;;
   purge|destroy)
     docker compose down --volumes --remove-orphans
-    [ "$MODE" = "destroy" ] && docker rmi ryanmillard/img2num-dev:latest
+    if [ "$MODE" = "destroy" ]; then
+      IMAGE_TO_REMOVE="$(load_state_image)"
+      docker rmi "$IMAGE_TO_REMOVE" 2>/dev/null || true
+      rm -f "$IMG2NUM_STATE_FILE"
+    fi
     ;;
 
   # Tail logs
--- a/img2num.ps1
+++ b/img2num.ps1
@@ -8,10 +8,38 @@ param(
     [string[]]$RemainingArgs
 )
 
+# Resolved image name for this invocation. Mirrors docker-compose.yml default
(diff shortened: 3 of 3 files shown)


Ryan-Millard (approved):
Thanks for the PR - really appreciate the contribution!🦔🦔

This is a solid improvement.

I’ve also extended this a bit further (it'll be in #315 I believe), so you’ll see more changes soon that build on what you started here.

Thanks again!

PR #335, opened 2026-04-29
feat: unified image_to_svg function as complete pipeline

### LINKED ISSUE #318: feat(core-api): consolidate into a single function
> [!TIP]
> This issue hasn't been claimed yet. Comment `/take` if you'd like to work on it!

---

### Describe the feature

This proposes allowing users to use a single function to invoke the entire pipeline we use in the example apps.

### It should be implemented because

It'll be a nice shorthand and abstract away unnecessary implementation details. Most good libraries do that!

### Additional context

Requirements:

1. Create a function called `image_to_svg` (the name is up for discussion if you can think of a better name) in the core library that accepts a raster image and applies the full pipeline to it. 
2. Create bindings for the new function in C, JavaScript, and Python (if #307 is merged)
3. Allow two version of the API
    - **Automatic setup**: pipeline created with the recommended parameters and functions.
    - **Manual setup**: consumers will pass something like a configuration struct to tell the program which functions to use to process the image (e.g., Gaussian blur vs bilateral filter) and parameters for each function.

### Would you like to work on this issue?

Yes

### LINKED ISSUE #360: refactor(bindings/js): enforce C-only source files and disable C++ in Emscripten bindings
> [!TIP]
> This issue hasn't been claimed yet. Comment `/take` if you'd like to work on it!

---

## Summary

The JS/Emscripten bindings in `bindings/js` are currently written as `.cpp` files with C++ compilation but must remain strictly C-compatible (extern "C" wrappers calling the C API). This creates a subtle footgun: developers may inadvertently use C++ features (e.g., default constructors, C++ structs) that are unavailable in pure C, which is the actual target ABI for the WASM binding layer.

## Motivation

Surfaced during code review in #335 ([comment](https://github.com/Ryan-Millard/Img2Num/pull/335#discussion_r3182231912)) — the use of `img2num_ImageToSvgConfig_default()` instead of a C++ default constructor was only noticed because the reviewer was aware of the C/C++ distinction. Future contributors may not have the same awareness.

## Proposed Changes

- Rename source files in `bindings/js/src/` from `.cpp` to `.c` (e.g., `wasm_wrapper.cpp` → `wasm_wrapper.c`).
- Update `bindings/js/CMakeLists.txt` to compile as C, not C++.
- Ensure the Emscripten toolchain is configured to reject C++ in this target.
- Audit all existing wrappers to confirm they use only C constructs (no C++ classes, references, templates, etc.).

## Requested by

@Ryan-Millard (PR #335)

### LINKED ISSUE #361: chore(py): add type annotations to public functions in api.py
> [!TIP]
> This issue hasn't been claimed yet. Comment `/take` if you'd like to work on it!

---

## Summary

All public functions in `packages/py/img2num/api.py` are missing type annotations. Adding annotations improves discoverability, IDE auto-complete, and enables downstream type checking.

Raised by @Ryan-Millard in a review comment on #335 (https://github.com/Ryan-Millard/Img2Num/pull/335#discussion_r3178303941).

## Affected functions

All functions below reside in `packages/py/img2num/api.py` and currently have no parameter or return type annotations:

| Function | Parameters (unannotated) | Expected return |
|---|---|---|
| `gaussian_blur_fft` | `image`, `sigma`, `width`, `height` | `numpy.ndarray` |
| `invert_image` | `image`, `width`, `height` | `numpy.ndarray` |
| `threshold_image` | `image`, `num_thresholds`, `width`, `height` | `numpy.ndarray` |
| `black_threshold_image` | `image`, `num_thresholds`, `width`, `height` | `numpy.ndarray` |
| `bilateral_filter` | `image`, `sigma_spatial`, `sigma_range`, `color_space`, `width`, `height` | `numpy.ndarray` |
| `kmeans` | `data`, `k`, `max_iter`, `color_space`, `width`, `height` | `numpy.ndarray` |
| `labels_to_svg` | `data`, `labels`, `min_area`, `width`, `height` | `str` |
| `image_to_svg` | `image`, `config`, `width`, `height` | `str` |

## Suggested approach

- Import `numpy.typing as npt` for array types (`npt.NDArray[np.uint8]`).
- Use `int`, `float`, and `str` for scalar and string types.
- Use `ImageToSvgConfig` (already imported) for the `config` parameter of `image_to_svg`.
- The keyword-only `width: int` and `height: int` parameters are injected by the `_inject_dims` decorator and should be annotated accordingly.
- Consider also annotating the `_inject_dims` helper with `typing.Callable` / `typing.TypeVar` for full coverage.

## References

- PR: #335
- Review comment: https://github.com/Ryan-Millard/Img2Num/pull/335#discussion_r3178303941
- Requester: @Ryan-Millard

Krasner:
Combines `bilateral_filter`, `kmeans` and `label_to_svg` into one unified `image_to_svg` function.
Updated js and py bindings

## What was changed & why

Fixes: #318 
Fixes: #360 
Fixes: #361


Krasner:
@Ryan-Millard overly paranoid lint check

example-apps/react-js/src/components/WasmImageProcessor.jsx
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useState, useId, useRef, useCallback, useMemo } from "react";
 import { Upload } from "lucide-react";
-import { imageToUint8ClampedArray, bilateralFilter, kmeans, findContours } from "img2num";
+import { imageToUint8ClampedArray, bilateralFilter, kmeans, findContours, imgToSVG } from "img2num";
```
Ryan-Millard: @Krasner, I agree that it is overly paranoid, but it's better to be pedantic than have chaos - especially if we are sending JS devs to look at this file to see how we did it.

This is just about the fact that you consolidated the pipeline into a single function, nothing else really.

```suggestion
import { imgToSVG } from "img2num";
```

packages/js/safeWasmWrappers.js
```diff
@@ -204,3 +204,38 @@ export const findContours = async ({ pixels, labels, width, height, min_area = 1
   });
   return { svg: result.returnValue, visualization: result.output.pixels };
 };
+
+/**
+ * @summary Convert images to SVG contours.
+ *
+ * @description
+ * Convert an input image into an SVG. A unification of bilateralFilter, kmeans, and findContours
```
Ryan-Millard:
```suggestion
 * @summary Convert raster images (e.g., JPEG, PNG) to SVGs.
 *
 * @description
 * Convert an input raster image into an SVG. A unification of `bilateralFilter`, `kmeans`, and `findContours`.
```

core/include/img2num.h
```diff
@@ -38,8 +38,14 @@ void bilateral_filter(uint8_t *image, size_t width, size_t height, double sigma_
...
 /// @copydoc IMG2NUM_H_LABELS_TO_SVG_DOC
-std::string labels_to_svg(const uint8_t *data, const int32_t *labels, const int width, const int height,
-                    const int min_area);
+std::string labels_to_svg(const uint8_t *data, const int32_t *labels, const int width,
+                          const int height, const int min_area);
+
+/// @copydoc IMG2NUM_H_IMAGE_TO_SVG_DOC
+std::string image_to_svg(const uint8_t *data, const int width, const int height,
+                         double sigma_spatial, double sigma_range, const int32_t k,
+                         const int32_t max_iter, const int min_area, const uint8_t color_space);
```
Ryan-Millard: What if we introduce a configuration struct here instead of keeping a long parameter list? This would make the API more extensible and reduce the risk of breaking changes in the future, since new parameters can be added without modifying the function signature.

Given that some parameters are logically grouped, it also makes sense for the config to be nested rather than flat. For example, clustering-related settings like `k` and `max_iter` clearly belong under a `kmeans` subgroup:

> It'll take me forever to get the below struct correct in this text box. I hope you get the rough idea.
```cpp
struct ImageToSvgConfig {
    struct BilateralFilterConfig {
      double sigma_spatial = 1.0;
      double sigma_range   = 1.0;
      uint8_t color_space  = 0
    } bilateral_filter;

    int min_area         = 50;;

    struct KMeansConfig {
        int32_t k        = 5;
        int32_t max_iter = 10;
    } kmeans;
};
```

Then the function becomes:

```cpp
std::string image_to_svg(
    const uint8_t *data,
    int width,
    int height,
    const ImageToSvgConfig& config = {}
);
```

Usage would look like:

```cpp
img2num::ImageToSvgConfig config = img2num::IMAGE_TO_SVG_DEFAULT_CONFIG;
config.kmeans.k = 7;
```

This approach improves clarity by grouping related parameters together, avoids ambiguity as the API grows, and keeps the function signature stable over time - we can just add new nested structs with default values.

core/src/internal/bilateral_filter.cpp
```diff
@@ -224,8 +225,10 @@ void bilateral_filter(uint8_t *image, size_t width, size_t height, double sigma_
     GPU::getClassInstance().init_gpu();
 
     if (GPU::getClassInstance().is_initialized()) {
+        std::cout << "bilateral_filter gpu" << std::endl;
         bilateral_filter_gpu(image, width, height, sigma_spatial, sigma_range, color_space);
     } else {
+        std::cout << "bilateral_filter cpu" << std::endl;
```
Ryan-Millard: Please remove the logging when you're done.

core/src/internal/bilateral_filter_gpu.cpp
```diff
@@ -286,7 +286,7 @@ void bilateral_filter_gpu(uint8_t* image, size_t width, size_t height, double si
     readBuffer.Destroy();
     delete waiting;
 #if defined(__EMSCRIPTEN__)
-    emscripten_sleep(50);
+    emscripten_sleep(100);
```
Ryan-Millard: Why did you change this?

core/src/internal/kmeans_gpu.cpp
```diff
@@ -665,6 +665,6 @@ void kmeans_gpu(const uint8_t* data, uint8_t* out_data, int32_t* out_labels, con
     labels.clear();
     labels.shrink_to_fit();
 #if defined(__EMSCRIPTEN__)
-    emscripten_sleep(50);
+    emscripten_sleep(100);
```
Ryan-Millard: Why is it now 100 and not 50?

example-apps/console-cpp/main.cpp
```diff
@@ -79,7 +81,7 @@ int main(int argc, char** argv) {
         exit_code = 1;
     }
     if (exit_code == 0) {
-        svgFile << res_svg;
+        svgFile << res_svg2;
         svgFile.close();
     }
```
Ryan-Millard, line 96: What is going on here?

example-apps/console-py/main.py
```diff
@@ -30,5 +30,10 @@ def main():
   with open(os.path.join(OUTDIR, "result.svg"),"w") as f:
       f.writelines(res_svg)
 
+  # res_svg2 should match res_svg
+  res_svg2 = img2num.image_to_svg(img, 3, 50, 16, 100, 100, 0)
+  with open(os.path.join(OUTDIR, "result2.svg"),"w") as f:
+      f.writelines(res_svg2)
+
```
Ryan-Millard, line 40: But why the duplication?

Ryan-Millard (changes requested):
This is very nice and clean. Thank you.

There are just a few minor questions I have and one major architectural discussion I think we need to have before this can be merged. The architectural concern is about the `core` `image_to_svg` function.

example-apps/console-py/main.py
Krasner: it's a comparison of the step wise approach versus the unified call

example-apps/console-cpp/main.cpp
Krasner: visualize the unified function output

core/src/internal/kmeans_gpu.cpp
Krasner: just an extra 50ms to help webgpu release memory. but it's not necessary. can change back

core/include/img2num.h
Krasner: how would this work with wasm?

Ryan-Millard: There are three main ways to do it. ChatGPT explains how it works [here](https://chatgpt.com/share/69f244b6-2998-83ea-b21f-51f3f391f583).

I don't really like the pointer or the Embind options because they're harder to maintain.

My idea is to _keep a nice nested struct in C++, but expose a flat ABI to JS and reconstruct it internally_.

We could basically do this:

**C++**
```cpp
struct ImageToSvgConfig {
    struct BilateralFilterConfig {
        double sigma_spatial;
        double sigma_range;
        uint8_t color_space;
    } bilateral_filter;

    int min_area;

    struct KMeansConfig {
        int32_t k;
        int32_t max_iter;
    } kmeans;
};

void process_impl(const ImageToSvgConfig& cfg) {
    // your real logic here
}

extern "C" {
    extern "C" {
    void process(
        double bilateralFilter_sigmaSpatial,
        double bilateralFilter_sigmaRange,
        uint8_t bilateralFilter_colorSpace,
        int min_area,
        int32_t kmeans_k,
        int32_t kmeans_max_iter
    ) {
        ImageToSvgConfig cfg;

        cfg.bilateral_filter.sigma_spatial = bilateralFilter_sigmaSpatial;
        cfg.bilateral_filter.sigma_range   = bilateralFilter_sigmaRange;
        cfg.bilateral_filter.color_space   = bilateralFilter_colorSpace;

        cfg.min_area = min_area;

        cfg.kmeans.k        = kmeans_k;
        cfg.kmeans.max_iter = kmeans_max_iter;

        process_impl(cfg);
    }
}
}
```

**JavaScript**
```js
const imgToSvg = async (width, height, config = {}) => {
  const bf = config.bilateral_filter ?? {};
  const km = config.kmeans ?? {};

  const result = await callWasm({
    funcName: "process",
    args: {
      bilateralFilter_sigmaSpatial: bf.sigma_spatial ?? 1.0,
      bilateralFilter_sigmaRange:   bf.sigma_range   ?? 1.0,
      bilateralFilter_colorSpace:   bf.color_space   ?? 0,
      min_area:                    config.min_area  ?? 50,
      kmeans_k:                    km.k             ?? 5,
      kmeans_max_iter:             km.max_iter      ?? 10,
    },
    bufferKeys: [],       // no pointers here
    returnType: "void",   // or "string" if you later return SVG
  });

  return result;
};
```

---

The C++ and JavaScript code will likely become ugly if the parameter list grows, but this "stupid" implementation will protect us from unnecessary overhead, overly complex code and future frustrations. The ugliness of the code isn't important since it will be internal and mostly ignored since the API shouldn't be changing often at all. If we later decide that we hate how we implemented it, we can always change it because we control the WASM-to-JS layer, not downstream consumers, which makes it very nice for everyone.

What are your thoughts on this?

example-apps/console-py/main.py
Ryan-Millard: Aren't the outputs identical, though?

Krasner: yes they should be nearly identical. only variation is caused by kmeans initialization randomness

core/include/img2num.h
Krasner: OK I implemented ImageToSvgConfig and all the bindings. Please review @Ryan-Millard

example-apps/react-js/src/components/WasmImageProcessor.jsx
Krasner: I do like the original step by step process better because it allows the hedgehog to move forward between steps. with imgToSVG all the steps happen in one call so the hedgehog leaps to the end

Ryan-Millard: That's a good point - I didn't think about that.

I think it would be better to revert the example app back to how it used to be. We should just make another example app for JS - I wanted that in any case: a Node.js console app to show that it works in the various runtimes.

---

@coderabbitai, please create an issue for this. We want a Node example app. It could be a simple MERN app, SSR app, whatever the person implementing it wants - as long as it imports Img2Num like the current Vite React app does and also uses the library server-side (not in the browser runtime like the React app does).

I recommend suggesting a cool app for the person implementing it to make so that it can properly demonstrate how the library can be used.

Ryan-Millard:
@Krasner, please see the review I previously left - some of those things are still relevant. Additionally, please clean up the old code that was commented out and update the C example app and ensure that the C bindings work (they currently cause the error below).

```sh
root@8c5abede0223:/usr/src/app# cmake --build build
No changes in /usr/src/app/build/embedded_shaders.h, skipping write
[  0%] Built target Img2Num_shaders
[ 68%] Built target Img2Num
[ 81%] Built target CImg2Num
[ 90%] Built target console_cpp_app
[ 95%] Building C object example-apps/console-c/CMakeFiles/console_c_app.dir/main.c.o
/usr/src/app/example-apps/console-c/main.c: In function 'main':
/usr/src/app/example-apps/console-c/main.c:55:5: error: variable 'config' has initializer but incomplete type
   55 |     ImageToSvgConfig config = {0};
      |     ^~~~~~~~~~~~~~~~
/usr/src/app/example-apps/console-c/main.c:55:32: warning: excess elements in struct initializer
   55 |     ImageToSvgConfig config = {0};
      |                                ^
/usr/src/app/example-apps/console-c/main.c:55:32: note: (near initialization for 'config')
/usr/src/app/example-apps/console-c/main.c:55:22: error: storage size of 'config' isn't known
   55 |     ImageToSvgConfig config = {0};
      |                      ^~~~~~
gmake[2]: *** [example-apps/console-c/CMakeFiles/console_c_app.dir/build.make:79: example-apps/console-c/CMakeFiles/console_c_app.dir/main.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:307: example-apps/console-c/CMakeFiles/console_c_app.dir/all] Error 2
gmake: *** [Makefile:136: all] Error 2
```

Krasner:
> update the C example app and ensure that the C bindings work

good catch. let me see what's going on there

Krasner:
> > update the C example app and ensure that the C bindings work
> 
> good catch. let me see what's going on there

OK fixed. Try compiling again @Ryan-Millard

Ryan-Millard:
Hi @Krasner. This is a lot to review and I'd rather do it when I have time to really think about what is going on.

I'll try to review it as soon as I have available time.

Krasner:
> Hi @Krasner. This is a lot to review and I'd rather do it when I have time to really think about what is going on.
> 
> I'll try to review it as soon as I have available time.

Take your time. The biggest change here is the `ImageToSvgConfig` struct which i think is actually pretty nice. But it does add some complexity to the bindings. The pybind implementation is the most "verbose" since it needs an implementation for each sub-struct. But overall not terrible.
Being able to do this in python is clean
```py
cfg = img2num.ImageToSvgConfig()
svg = img2num.image_to_svg(img, cfg)
```

Krasner:
right now the way the pybinding works for `img2num.ImageToSvgConfig`:
```py
>>> cfg = img2num.ImageToSvgConfig()
>>> cfg # prints out __repr__
<ImageToSvgConfig {bilateral_filter: {'sigma_spatial': 3.000000, 'sigma_range': 50.000000}, min_cluster_area: 100, color_space: 0, kmeans: {'k': 16, 'max_iter': 100}}>

>>> cfg.bilateral_filter.sigma_spatial = 2.0 // modify values
>>> cfg
<ImageToSvgConfig {bilateral_filter: {'sigma_spatial': 2.000000, 'sigma_range': 50.000000}, min_cluster_area: 100, color_space: 0, kmeans: {'k': 16, 'max_iter': 100}}>

>>> cfg.kmeans.k=32
>>> cfg
<ImageToSvgConfig {bilateral_filter: {'sigma_spatial': 2.000000, 'sigma_range': 50.000000}, min_cluster_area: 100, color_space: 0, kmeans: {'k': 32, 'max_iter': 100}}>
```
but I don't have a different constructor. So `cfg = img2num.ImageToSvgConfig(spatial_sigma=3.0, sigma_range=50.0)` would not work. 
maybe worth implementing

bindings/c/src/cimg2num.cpp
```diff
@@ -55,4 +55,36 @@ char *img2num_labels_to_svg(const uint8_t *data, const int32_t *labels, const in
...
+    img2num::clear_last_error_and_catch(
+        [&](const uint8_t *d, const int w, const int h, const ImageToSvgConfig* cfg) {
+            std::string svg{img2num::image_to_svg(d, w, h, *cfg)};
+            result = static_cast<char *>(std::malloc(svg.size() + 1));
+            if (!result) {
+                return;  // Allocation failed
+            }
+            std::memcpy(result, svg.c_str(), svg.size() + 1);
+        },
+        data, width, height, config);
```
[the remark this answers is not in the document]
Ryan-Millard: @coderabbitai

If this is outdated, please resolve it or give an updated comment about this.

core/include/img2num.h
```diff
@@ -15,6 +15,28 @@
...
+        int32_t max_iter = 100;
+    } kmeans;
+};
+
+const ImageToSvgConfig IMAGE_TO_SVG_DEFAULT_CONFIG = {
+    .bilateral_filter{.sigma_spatial = 3.0, .sigma_range=50.0},
+    .min_cluster_area = 100,
+    .color_space = 0,
+    .kmeans{.k=16, .max_iter=100}
+};
```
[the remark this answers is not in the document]
Ryan-Millard: @coderabbitai

If this is outdated, please resolve it or give an updated comment about this.

core/src/internal/image_to_svg.cpp
```diff
@@ -0,0 +1,25 @@
...
+//                         double sigma_spatial, double sigma_range, const int32_t k,
+//                         const int32_t max_iter, const int min_area, const uint8_t color_space) {
+std::string image_to_svg(const uint8_t *data, const int width, const int height, const ImageToSvgConfig& config) {
+    // self deallocate
+    std::vector<uint8_t> img_data(width * height * 4);
+    std::vector<uint8_t> out_data(width * height * 4);
+    std::vector<int32_t> out_labels(width * height);
+
+    std::memcpy(img_data.data(), data,
+                static_cast<size_t>(width) * static_cast<size_t>(height) * 4);
```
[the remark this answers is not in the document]
Ryan-Millard: @coderabbitai

If this is outdated, please resolve it or give an updated comment about this.

bindings/c/include/cimg2num.h
```diff
@@ -13,9 +13,29 @@
...
+    } bilateral_filter;
+    int min_cluster_area;
+    uint8_t color_space;
+    struct {
+        int32_t k;
+        int32_t max_iter;
+    } kmeans;
+} ImageToSvgConfig;
+
 #endif
```
Ryan-Millard: ## Overview

This makes the header confusing and will likely lead to errors for some projects. The new struct should also take the C-style namespace (a name prefix: `img2num_`) to avoid naming conflicts.

### Header Confusion

<details>
<summary>Details</summary>

This exposes two versions of `ImageToSvgConfig`:

1.  `ImageToSvgConfig` for C (the people who should be using this header).
2. `img2num::ImageToSvgConfig` for C++ (the people who should rather use `img2num.h`)

This is dangerous because:

1. It doesn't clearly show people which struct to use, meaning that they will need to check again every time they change languages.
2. It is very reliant on the environment that is using the header.

**My opinion:**
I think that we should force this to be a C-style header. We only have the C++ check to ensure that the code is compiled in a C-friendly manner if a C++ compiler is used. The `#ifdef __cplusplus` is only there to prevent it from being in the header when a C compiler is used because C doesn't have `extern "C"` because that would make sense (C is inherently compatible with itself).

</details>

### Compilation / Type Errors

<details>
<summary>Details</summary>

Anyone who uses this header (`cimg2num.h`) expects the code inside of it to be self-compatible. However, this is not the case when a C++ compiler crosses it because it will have `typedef img2num::ImageToSvgConfig ImageToSvgConfig;` in the header instead of the C struct. That sounds correct until we try to use the C++ struct with the C function that consumes it - `img2num_image_to_svg` expects a completely different type: the C struct. 

**My opinion**
No C++ code should exist in this file because it is for C consumers. If a C++ user wants to use our library, they must:

1. Use the header designed explicitly for them: `img2num.h`
2. Write C-style code using the C header: `cimg2num.h`

If a C++ user tries to use `cimg2num.h` like it is a C++ header, they need to get into trouble. Our job is to make clean code and document how to use, not to ensure that people who use it incorrectly don't pay the price for doing so.


</details>

---

## Suggestions

1. Remove the include of `img2num.h` because this is a C header and C++ users are not first-class citizens in this context.
2. Remove the C++ struct definition for the same reason as point 1.
3. Move the struct out of the macro conditional so it is accessible no matter which environment compiles it.
4. Put the struct in the appropriate C-style namespace: `img2num_`

> Like this, we only add `extern "C"` when in a C++ environment so that the code is compiled without C++ things (like name mangling).


```suggestion
#ifdef __cplusplus
extern "C" {
#endif

typedef struct img2num_ImageToSvgConfig {
    struct {
        double sigma_spatial;
        double sigma_range;
    } bilateral_filter;
    int min_cluster_area;
    uint8_t color_space;
    struct {
        int32_t k;
        int32_t max_iter;
    } kmeans;
} ImageToSvgConfig;
```

core/include/img2num.h
```diff
@@ -15,6 +15,27 @@
...
+        int32_t k = 16;
+        int32_t max_iter = 100;
+    } kmeans;
+};
+
+const ImageToSvgConfig IMAGE_TO_SVG_DEFAULT_CONFIG = {
+    .bilateral_filter{.sigma_spatial = 3.0, .sigma_range = 50.0},
+    .min_cluster_area = 100,
+    .color_space = 0,
+    .kmeans{.k = 16, .max_iter = 100}};
```
Ryan-Millard: We don't actually need the default config instance since every instance of this struct will have those values by default. A user who wants the defaults can just restrain themselves from altering them.😁

Just remember that this will break the recommendation I made for `cimg2num.h` since it relies on the default from here. That's okay - just create an instance of this struct in the source file (`cimg2num.cpp`) and use that instead of the default we delete here because a `.cpp` is always compiled in a C++ environment.


```suggestion
struct ImageToSvgConfig {
    struct BilateralFilterConfig {
        double sigma_spatial = 3.0;
        double sigma_range = 50.0;
    } bilateral_filter;

    int min_cluster_area = 100;
    uint8_t color_space = 0;

    struct KMeansConfig {
        int32_t k = 16;
        int32_t max_iter = 100;
    } kmeans;
};
```

Ryan-Millard (changes requested):
I'm unfortunately ending this review early for today, but wanted to give you some feedback so you can understand my dilemma with this code. I'll have to come back and look at it when my head isn't as foggy and I've researched it a bit more. This is a fairly complex and uncommon thing to tackle, so I'd like to remind myself of how to do it.

I am concerned about the management of the struct between the C++ and C code.

I think the closest similar implementation we have to this is in [`bindings/c/include/img2num_error_t.h`](https://github.com/Ryan-Millard/Img2Num/blob/main/bindings/c/include/cimg2num/img2num_error_t.h) (also see [the source file](https://github.com/Ryan-Millard/Img2Num/blob/main/bindings/c/src/cimg2num/img2num_error_t.cpp)) and [`core/include/img2num/Error.h`](https://github.com/Ryan-Millard/Img2Num/blob/main/core/include/img2num/Error.h) (also see [`core/src/img2num/Error.cpp`](https://github.com/Ryan-Millard/Img2Num/blob/main/core/src/img2num/Error.cpp)). My aim with those error files was to protect against exceptions - which are unknown in C contexts and unfriendly in certain C++ codebases - by returning error codes instead.

I hope we'll be able to find a clean solution to our problem that doesn't require a lot of manual work when the API updates.

bindings/c/include/cimg2num.h
Krasner: This is not a good solution because it doesn't directly map the C++ struct with the C struct. With your proposal I will have to explicitly use. I guess it's fine but ugly.

```
reinterpret_cast<img2num_ImageToSvgConfig*>(new img2num::ImageToSvgConfig);
```
so these functions change:
```cpp
ImageToSvgConfig* img2num_config_create() {
    return reinterpret_cast<img2num_ImageToSvgConfig*>(new img2num::ImageToSvgConfig);
}

char *img2num_image_to_svg(const uint8_t *data, const int width, const int height,
                           const ImageToSvgConfig *config) {
    char *result{nullptr};
    img2num::clear_last_error_and_catch(
        [&](const uint8_t *d, const int w, const int h, const ImageToSvgConfig *cfg) {
            std::string svg{img2num::image_to_svg(d, w, h, *reinterpret_cast<const img2num::ImageToSvgConfig*>(cfg))};
            result = static_cast<char *>(std::malloc(svg.size() + 1));
            if (!result) {
                return;  // Allocation failed
            }
            std::memcpy(result, svg.c_str(), svg.size() + 1);
        },
        data, width, height, config);
    return result;
};
```

Thoughts?

Krasner: It is annoying to have to redefine the structs in both C++ and C implementations explicitly...

Krasner: @Ryan-Millard let me know what you think here. I'm also thinking about separating out the BilateralFilter config and the KMeans config:

in `img2num.h`:
```cpp
struct BilateralFilterConfig {
    double sigma_spatial = 3.0;
    double sigma_range = 50.0;
};

struct KMeansConfig {
    int32_t k = 16;
    int32_t max_iter = 100;
};

struct ImageToSvgConfig {
    BilateralFilterConfig bilateral_filter;
    KMeansConfig kmeans;
    int min_cluster_area = 100;
    uint8_t color_space = 0;
};
```

in `cimg2num.h`:
```c
#ifdef __cplusplus
extern "C" {
#endif
typedef struct img2num_BilateralFilterConfig {
    double sigma_spatial;
    double sigma_range;
} BilateralFilterConfig;

typedef struct img2num_KMeansConfig {
    int32_t k;
    int32_t max_iter;
} KMeansConfig;

typedef struct img2num_ImageToSvgConfig {
    BilateralFilterConfig bilateral_filter;
    KMeansConfig kmeans;
    int min_cluster_area;
    uint8_t color_space;
} ImageToSvgConfig;
```

Ryan-Millard: @Krasner, we can separate them, but what actual value will that provide us with? Our functions are already quite simple and if we use BilateralFilterConfig and KMeansConfig, it'll just make things more confusing.

Imagine this:

```cpp
struct AddConfig {
    int a;
    int b;
}

int add(AddConfig cfg) {
    return cfg.a + cfg.b;
}

int main() {
    AddConfig cfg{ .a = 5, .b = 6 };
    add(cfg);
}
```

It may be nicer for us, but we still need to think about the people who will use the library one day. It would be very weird to do that.

The reason I suggested a nested struct is because users won't always want to pass every value - it lets us create defaults and it lets them have compile-time guarantees that they used the library properly:

```cpp
int main() {
     uint8_t* data = ...;
    int width = 250, height = 250;   
    img2num::ImageToSvgConfig cfg{}; // default values
    cfg.kmeans.k = 16; // simply update one config value

    img2num::image_to_svg(data, width, height, cfg);
}
```

This also prevents having so many arguments that the screen can't fit all of them in it.

Krasner: makes sense. not hard to separate them out later if need be

Krasner: @Ryan-Millard in your latest (i think deleted comment) the `to_c` / `to_cpp` functions are essentially what `reinterepret_cast` is already doing

Ryan-Millard:
@Krasner, I've done a lot of looking around and it seems there are two good ways to do it as I show below. I'd like your opinion on it.

<details><summary>I put this in a details block so we can scroll through this chat more easily</summary>
<p>


The main driver behind this approach is UX. C users shouldn't have to manage heap memory just to configure a function call - requiring `img2num_config_create` / `img2num_config_free` adds cognitive overhead, introduces a resource leak foothole, and breaks the expectation that you can free() anything heap-allocated from a C library. A plain value type lets users do:

In `img2num.h`:
```cpp
namespace img2num {
    struct ImageToSvgConfig {
        struct BilateralFilterConfig {
            double sigma_spatial = 3.0;
            double sigma_range = 50.0;
        } bilateral_filter;
    
        int min_cluster_area = 100;
        uint8_t color_space = 0;
    
        struct KMeansConfig {
            int32_t k = 16;
            int32_t max_iter = 100;
        } kmeans;
    };
}
```

In `cimg2num.h`:
```c
typedef struct {
    struct {
        double sigma_spatial;
        double sigma_range;
    } bilateral_filter;

    int min_cluster_area;
    uint8_t color_space;

    struct {
        int32_t k;
        int32_t max_iter;
    } kmeans;
} img2num_ImageToSvgConfig;

img2num_ImageToSvgConfig img2num_ImageToSvgConfig_default(void);
```

In `cimg2num.cpp`:
```cpp
// =================================
// UTILS
// =================================

static img2num::ImageToSvgConfig to_cpp(const img2num_ImageToSvgConfig& c) {
    return {
        .bilateral_filter{
            .sigma_spatial = c.bilateral_filter.sigma_spatial,
            .sigma_range   = c.bilateral_filter.sigma_range
        },
        .min_cluster_area = c.min_cluster_area,
        .color_space      = c.color_space,
        .kmeans{
            .k        = c.kmeans.k,
            .max_iter = c.kmeans.max_iter
        }
    };
}

static img2num_ImageToSvgConfig to_c(const img2num::ImageToSvgConfig& cpp) {
    return {
        .bilateral_filter{
            .sigma_spatial = cpp.bilateral_filter.sigma_spatial,
            .sigma_range   = cpp.bilateral_filter.sigma_range
        },
        .min_cluster_area = cpp.min_cluster_area,
        .color_space      = cpp.color_space,
        .kmeans{
            .k        = cpp.kmeans.k,
            .max_iter = cpp.kmeans.max_iter
        }
    };
}

img2num_ImageToSvgConfig img2num_ImageToSvgConfig_default(void) {
    // Use the defaults from the C++ struct
    return to_c(img2num::ImageToSvgConfig{});
}

// =================================
// BINDINGS
// =================================


char *img2num_image_to_svg(const uint8_t *data, const int width, const int height,
    const img2num::ImageToSvgConfig config{to_cpp(*c_config)};
    // The rest of the function
};

```

---

I personally prefer the approach that I just suggested because it make the library the simplest for anyone using it (and that's what matters most). People generally turn away when a library is difficult to use, so making it simpler would increse the number of people that actually want to use (and maybe even contribute to) the library.

**Your approach (heap-allocated, pointer):**
- `img2num_config_create` / `img2num_config_free`
- C users get default values easily, but must remember to call `img2num_config_free`, can't use free(), and it's awkward to pass around. I have no idea what would happen if someone tried to use `free()` on that struct.

My approach (value type, C struct):
- Constantly converting between C-style and C++-style structs
- Users just declare `img2num_ImageToSvgConfig cfg = img2num_ImageToSvgConfig_default();` and pass &cfg — no memory management, free()-safe, stack-allocated.
- The cost is the to_cpp/to_c conversion layer and the duplicated struct.

This doesn't mean that your approach is bad - it's actually really nice. The only problem is that it optimizes for something I think we should just suck up and deal with so we can produce a user-friendly library.

This leaves us with two options:
1. Good user experience, but bad developer experience (my value-type version)
2. Good developer experience, but bad user experience (your pointer-based version)

I don't like either, but I prefer the the value-type one because we can just keep good documentation for it - it'll be frustrating to deal with sometimes, but save us from people opening the same issues all the time. External developers might also hate having to read documentation or not even know about it.




</p>
</details>

bindings/c/include/cimg2num.h
Ryan-Millard: `reinterpret_cast` is only guaranteed to be safe in C++. Unfortunately, since C and C++ have different memory systems (e.g., C++ pads structs and C doesn't), this won't always guarantee that the memory will be aligned properly for C to read it.

It may work well right now because the struct is simple, but it might lead to problems later on if we modify the C++ struct.

Krasner:
> @Krasner, I've done a lot of looking around and it seems like the best way to do it is the way I show below.
> 
> I put this in a details block so we can scroll through this chat more easily

OK this makes sense... it's incredibly verbose but is easier to understand. Take a look at the latest commit

bindings/c/include/cimg2num.h
```diff
@@ -16,6 +16,23 @@
 extern "C" {
 #endif
 
+// mirrors struct definition in img2num.h
+
+typedef struct img2num_ImageToSvgConfig {
```
Ryan-Millard: Please add a Doxygen comment above this and make some API comments in `doxygen/img2num.h.dox`, then reuse it for the `img2num.h` C++ version.

This will make it simpler for new people to use our library.

bindings/c/include/cimg2num.h
```diff
@@ -43,6 +60,9 @@ void img2num_bilateral_filter(uint8_t *image, size_t width, size_t height, doubl
 char *img2num_labels_to_svg(const uint8_t *data, const int32_t *labels, const int width,
                             const int height, const int min_area);
 
+/// @copydoc ::IMG2NUM_H_IMAGE_TO_SVG_DOC
```
Ryan-Millard, line 81: Please add this Doxygen comment in `doxygen/img2num.h.dox`.

Doxygen is a bit whack because it treats these as macros, so naming them consistently is important. You did that, but I just wanted to let you know about that strange behavior in case it ever bites you.

bindings/c/src/cimg2num.cpp
```diff
@@ -7,6 +7,27 @@
...
+            .color_space = c.color_space};
+}
+
+static img2num_ImageToSvgConfig to_c(const img2num::ImageToSvgConfig &cpp) {
+    return {.bilateral_filter{.sigma_spatial = cpp.bilateral_filter.sigma_spatial,
+                              .sigma_range = cpp.bilateral_filter.sigma_range},
+            .kmeans{.k = cpp.kmeans.k, .max_iter = cpp.kmeans.max_iter},
+            .min_cluster_area = cpp.min_cluster_area,
+            .color_space = cpp.color_space};
+}
```
Ryan-Millard, line 38: Just a formatting fix for readability.

```suggestion
static img2num::ImageToSvgConfig to_cpp(const img2num_ImageToSvgConfig &c)
{
    return {
        .bilateral_filter{
            .sigma_spatial = c.bilateral_filter.sigma_spatial,
            .sigma_range   = c.bilateral_filter.sigma_range
        },
        .kmeans{
            .k        = c.kmeans.k,
            .max_iter = c.kmeans.max_iter
        },
        .min_cluster_area = c.min_cluster_area,
        .color_space      = c.color_space
    };
}

static img2num_ImageToSvgConfig to_c(const img2num::ImageToSvgConfig &cpp)
{
    return {
        .bilateral_filter{
            .sigma_spatial = cpp.bilateral_filter.sigma_spatial,
            .sigma_range   = cpp.bilateral_filter.sigma_range
        },
        .kmeans{
            .k        = cpp.kmeans.k,
            .max_iter = cpp.kmeans.max_iter
        },
        .min_cluster_area = cpp.min_cluster_area,
        .color_space      = cpp.color_space
    };
}
```

bindings/js/src/wasm_wrapper.c
```diff
@@ -37,4 +37,22 @@ EMSCRIPTEN_KEEPALIVE char *labels_to_svg(uint8_t *data, int32_t *labels, const i
...
+
+    config.bilateral_filter.sigma_spatial = sigma_spatial;
+    config.bilateral_filter.sigma_range = sigma_range;
+    config.kmeans.k = k;
+    config.kmeans.max_iter = max_iter;
+    config.min_cluster_area = min_area;
+    config.color_space = color_space;
+
+    return img2num_image_to_svg(data, width, height, &config);
+}
```
Ryan-Millard, line 54: Just cleanup.

```suggestion
EMSCRIPTEN_KEEPALIVE char *image_to_svg(const uint8_t *data, const int width, const int height,
                                        double sigma_spatial, double sigma_range, const int32_t k,
                                        const int32_t max_iter, const int min_area,
                                        const uint8_t color_space) {
    img2num_ImageToSvgConfig config = img2num_ImageToSvgConfig_default();

    config.bilateral_filter.sigma_spatial = sigma_spatial;
    config.bilateral_filter.sigma_range = sigma_range;
    config.kmeans.k = k;
    config.kmeans.max_iter = max_iter;
    config.min_cluster_area = min_area;
    config.color_space = color_space;

    return img2num_image_to_svg(data, width, height, &config);
}
```

This confused me at first because I was wondering why you used the default function here instead of just letting C++ do it, then I realized that the Emscripten bindings are in C, not C++. We need to update it all to force us to use C there so we can't shoot ourselves in the feet later. 

@coderabbitai please make an issue for this (related to making `bindings/js` C-only and disable C++) and label it as important.

bindings/js/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ file(GLOB_RECURSE BINDING_SRC CONFIGURE_DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/src
 
 add_executable(img2num_wasm ${BINDING_SRC})
 
-target_link_libraries(img2num_wasm PRIVATE CImg2Num)
+target_link_libraries(img2num_wasm PRIVATE CImg2Num Img2Num)
```
Ryan-Millard: This can be reverted. `Img2Num` is linked to this through`CImg2Num` since `CImg2Num` relies on it. `bindings/js` doesn't rely directly on `Img2Num` (`core`).

```suggestion
target_link_libraries(img2num_wasm PRIVATE CImg2Num)
```

bindings/py/src/img2num_pybind.cpp
```diff
@@ -120,4 +121,93 @@ PYBIND11_MODULE(_img2num, m) {
...
+                 cfg = img2num.ImageToSvgConfig(
+                     bf = {"sigma_spatial": 5.0, "sigma_range": 30.0},
+                     km = {"k": 32},
+                     min_cluster_area = 250,
+                     color_space = 1
+                 )
+
+                 cfg = img2num.ImageToSvgConfig() will use default values
+                 */
+
```
Ryan-Millard:
```suggestion
```

bindings/py/src/img2num_pybind.cpp
```diff
@@ -3,9 +3,10 @@
 #include <pybind11/pybind11.h>
 
 #include <cstdlib>
+#include <memory>
```
Ryan-Millard, line 7: For safety, we need this new include for `std::string_stream`. It's probably included via another include, but this is safer.

```suggestion
#include <img2num.h>

#include <pybind11/numpy.h>
#include <pybind11/pybind11.h>

#include <cstdlib>
#include <memory>
#include <sstream>
```

bindings/py/src/img2num_pybind.cpp
```diff
@@ -120,4 +121,93 @@ PYBIND11_MODULE(_img2num, m) {
...
+                 // 3. Process KMeans overrides from the 'km' dictionary
+                 if (km_dict.contains("k")) c->kmeans.k = km_dict["k"].cast<int>();
+                 if (km_dict.contains("max_iter"))
+                     c->kmeans.max_iter = km_dict["max_iter"].cast<int>();
+
+                 // 4. Process remaining top-level kwargs (like color_space or min_cluster_area)
+                 if (kwargs.contains("min_cluster_area"))
+                     c->min_cluster_area = kwargs["min_cluster_area"].cast<int>();
+                 if (kwargs.contains("color_space"))
+                     c->color_space = kwargs["color_space"].cast<uint8_t>();
```
Ryan-Millard, line 170: It's often easy to make mistakes with these, so I typically prefer to use curly braces - even if it's for a one-liner.

The choice is up to you, though. This is your code and I'd like it to look the way you want it to look.

core/include/img2num.h
```diff
@@ -15,6 +15,19 @@
...
+        double sigma_spatial = 3.0;
+        double sigma_range = 50.0;
+    } bilateral_filter;
+    struct KMeansConfig {
+        int32_t k = 16;
+        int32_t max_iter = 100;
+    } kmeans;
+    int min_cluster_area = 100;
+    uint8_t color_space = 0;
+};
```
Ryan-Millard, line 49: Please add a Doxygen DocString for this. It is very necessary because it will allow users to understand how to use our library.

core/include/img2num.h
```diff
@@ -40,6 +53,11 @@ void bilateral_filter(uint8_t *image, size_t width, size_t height, double sigma_
 /// @copydoc IMG2NUM_H_LABELS_TO_SVG_DOC
 std::string labels_to_svg(const uint8_t *data, const int32_t *labels, const int width,
                           const int height, const int min_area);
+
+/// @copydoc IMG2NUM_H_IMAGE_TO_SVG_DOC
```
Ryan-Millard, line 77: Please add this to `doxygen/img2num.h.dox`. It is very important.

It isn't there yet:
```sh
ryan@Ryans-PC:~/projects/Img2Num$ grep "IMG2NUM_H_IMAGE_TO_SVG_DOC" doxygen/img2num.h.dox
ryan@Ryans-PC:~/projects/Img2Num$
```

core/src/internal/image_to_svg.cpp
```diff
@@ -0,0 +1,24 @@
...
+    std::vector<uint8_t> img_data(static_cast<size_t>(width) * static_cast<size_t>(height) * 4);
+    std::vector<uint8_t> out_data(static_cast<size_t>(width) * static_cast<size_t>(height) * 4);
+    std::vector<int32_t> out_labels(static_cast<size_t>(width) * static_cast<size_t>(height));
+
+    std::memcpy(img_data.data(), data,
+                static_cast<size_t>(width) * static_cast<size_t>(height) * 4);
+    bilateral_filter(img_data.data(), width, height, config.bilateral_filter.sigma_spatial,
+                     config.bilateral_filter.sigma_range, config.color_space);
+    kmeans(img_data.data(), out_data.data(), out_labels.data(), width, height, config.kmeans.k,
+           config.kmeans.max_iter, config.color_space);
```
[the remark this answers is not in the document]
Ryan-Millard: @coderabbitai please show what your proposed setup for this would look like. I understand the good intentions of #318, but I don't know how feasible it will actually be. UX is very important and I'm worried that this will become overengineered garbage if we do that.

example-apps/console-c/main.c
Ryan-Millard, line 1: I still don't quite understand the purpose of `res_svg2`. You said that it shows that the outputs will be different, but does that really need to be part of the example app?

I'm not saying it's a bad idea - I'd just like to understand it before I can comment on it.

example-apps/console-cpp/main.cpp
```diff
@@ -66,6 +66,10 @@ int main(int argc, char** argv) {
     // Generate SVG
     std::string res_svg{img2num::labels_to_svg(img_data, out_labels, width, height, 100)};
 
+    img2num::ImageToSvgConfig config;
+    config.kmeans.k = 32;
+    std::string res_svg2{img2num::image_to_svg(img_data, width, height, config)};
```
Ryan-Millard, line 71: @coderabbitai said this:

_⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_

**Use the original image buffer for `image_to_svg` in this example.**

`img_data` is already modified by the manual bilateral/kmeans path, so this no longer demonstrates the unified API on raw input as intended.

<details>
<summary>Proposed fix</summary>

```diff
-    std::string res_svg2{img2num::image_to_svg(img_data, width, height, config)};
+    std::string res_svg2{img2num::image_to_svg(image_data_original, width, height, config)};
```
</details>

 
As per coding guidelines, `example-apps/**` should correctly reflect public API usage with clean, minimal examples.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against the current code and only fix it if needed.

In `@example-apps/console-cpp/main.cpp` at line 69, The call to
img2num::image_to_svg is using img_data which has been modified by the manual
bilateral/kmeans path; change it to use the original unmodified image buffer
(e.g., save the raw input into a separate variable before applying
bilateral/kmeans and pass that raw buffer to img2num::image_to_svg) so the
example demonstrates the unified API on raw input; ensure references to
image_to_svg and the bilateral/kmeans modification code use the preserved
original buffer variable rather than the mutated img_data.
```

</details>

example-apps/console-py/main.py
```diff
@@ -30,5 +30,20 @@ def main():
...
+  # res_svg2 should match res_svg
+  cfg = img2num.ImageToSvgConfig(km = {"k": 16})
+  print(cfg)
+
+  # cfg.bilateral_filter.sigma_spatial = 3
+  # cfg.bilateral_filter.sigma_range = 50
+  # cfg.kmeans.k = 16
+  # cfg.kmeans.max_iter = 100
+  # cfg.min_cluster_size = 100
+  # cfg.color_space = 0
```
Ryan-Millard:
```suggestion
```

example-apps/react-js/src/components/WasmImageProcessor.jsx
```diff
@@ -96,7 +103,7 @@ const WasmImageProcessor = () => {
         labels,
         width,
         height,
-      });
+      });*/
```
Ryan-Millard:
```suggestion
```

packages/js/safeWasmWrappers.js
```diff
@@ -204,3 +204,39 @@ export const findContours = async ({ pixels, labels, width, height, min_area = 1
...
+ */
+export const imgToSVG = async ({ pixels, width, height, sigma_spatial = 3, sigma_range = 50, num_colors = 16, max_iter = 100, min_area = 100, color_space = 0 }) => {
+  const result = await callWasm({
+    funcName: "image_to_svg",
+    args: { pixels, width, height, sigma_spatial, sigma_range, num_colors, max_iter, min_area, color_space },
+    bufferKeys: [{ key: "pixels", type: "Uint8ClampedArray" }],
+    returnType: "string",
+  });
+  return { svg: result.returnValue, visualization: result.output.pixels };
+};
```
Ryan-Millard, line 240: This fixes the JSDoc comment. 

Additionally, shouldn't this function be called `imageToSvg` to match the names we gave it in the other languages?

```suggestion
/**
 * @summary Convert raster images (e.g., JPEG, PNG) to SVGs.
 *
 * @description
 * Convert an input raster image into an SVG. A unification of `bilateralFilter`, `kmeans`, and `findContours`.
 *
 * @async
 * @function imgToSVG
 * @param {Object} __named_parameters - The input options.
 * @property {Uint8ClampedArray} __named_parameters.pixels - Original image pixels.
 * @property {number} __named_parameters.width - Image width.
 * @property {number} __named_parameters.height - Image height.
 * @property {number} [__named_parameters.sigma_spatial=3] - Spatial standard deviation.
 * @property {number} [__named_parameters.sigma_range=50] - Range (color) standard deviation.
 * @property {number} [__named_parameters.num_colors=16] - Number of color clusters.
 * @property {number} [__named_parameters.max_iter=100] - Maximum number of iterations.
 * @property {number} [__named_parameters.min_area=100] - Minimum area of a region to be considered a contour.
 * @property {number} [__named_parameters.color_space=0] - Color space mode.
 * @returns {Promise<{svg: string, visualization: Uint8ClampedArray}>} Generated SVG and optionally pixels with visualized contours.
 * @throws {Error} If the WASM function fails or input labels are invalid.
 * @example
 * const { svg } = await imgToSVG({ pixels, width, height });
 * @todo Add support for constructor-style config input once bindings expose it.
 * @variation Unified raster-to-SVG pipeline with configurable filtering and clustering
 * @since 0.0.0
 */
export const imgToSVG = async ({ pixels, width, height, sigma_spatial = 3, sigma_range = 50, num_colors = 16, max_iter = 100, min_area = 100, color_space = 0 }) => {
  const result = await callWasm({
    funcName: "image_to_svg",
    args: { pixels, width, height, sigma_spatial, sigma_range, num_colors, max_iter, min_area, color_space },
    bufferKeys: [{ key: "pixels", type: "Uint8ClampedArray" }],
    returnType: "string",
  });
  return { svg: result.returnValue, visualization: result.output.pixels };
};
```

packages/py/img2num/api.py
```diff
@@ -9,6 +9,8 @@
     bilateral_filter        as _bilateral_filter,
     kmeans                  as _kmeans,
     labels_to_svg           as _labels_to_svg,
+    image_to_svg            as _image_to_svg,
+    ImageToSvgConfig
```
Ryan-Millard, line 15: `ImageToSvgConfig` was imported but never used.

packages/py/img2num/api.py
```diff
@@ -59,3 +63,7 @@ def kmeans(data, k, max_iter, color_space, *, width, height):
 @_inject_dims("data")
 def labels_to_svg(data, labels, min_area, *, width, height):
     return _labels_to_svg(data, labels, width, height, min_area)
+
+@_inject_dims("image")
+def image_to_svg(image, config, *, width, height):
+    return _image_to_svg(image, width, height, config)
```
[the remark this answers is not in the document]
Ryan-Millard: @coderabbitai many of our functions for Python do not have them. Please create an issue to add type annotations to these functions.

README.md
```diff
@@ -10,7 +10,7 @@
 [![Open in Codespaces](https://img.shields.io/badge/-Open%20in%20Codespaces-black?logo=github)](https://codespaces.new/Ryan-Millard/Img2Num)
 
 <a href="https://skillicons.dev">
-  <img height="25px" src="https://skillicons.dev/icons?i=cpp,c,js" />
+  <img height="25px" src="https://skillicons.dev/icons?i=cpp,c,js,py" />
```
Ryan-Millard:
```suggestion
  <img height="25px" src="https://skillicons.dev/icons?i=cpp,c,js,py" alt="C++, C, JavaScript, Python" />
```

Ryan-Millard (changes requested):
Nice work @Krasner!

I really like this. I'm so glad that I have someone else to help maintain this project. I appreciate the work you've put into it.

Ryan-Millard:
> > @Krasner, I've done a lot of looking around and it seems like the best way to do it is the way I show below.
> > I put this in a details block so we can scroll through this chat more easily
> 
> OK this makes sense... it's incredibly verbose but is easier to understand. Take a look at the latest commit

I agree 100%, but that's just C and C++ - you can't have nice things when using either of the two.

packages/py/img2num/api.py
Krasner: It's necessary to do:
```
import img2num
cfg = img2num.ImageToSvgConfig()
```
if it's not explicitly exposed like so, it `ImageToSvgConfig` will be missing

core/src/internal/image_to_svg.cpp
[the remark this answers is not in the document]
Ryan-Millard: This can be done some other time. The Gaussian blur is still problematic because of its zero-padding - it adds a vignette-like effect on the edges of the image.

packages/py/img2num/api.py
Ryan-Millard: Okay, I understand that it is just re-exported. 

Would it be possible to make this an optional parameter for `image_to_svg` so that they don't have to configure it if they want the default values applied?

Krasner: yes

example-apps/console-c/main.c
Krasner: It's only to show and verify that one can call the pipeline explicitly in steps (res_svg) or in the unified call (res_svg2) and get the same results

example-apps/console-py/main.py
```diff
@@ -30,5 +30,13 @@ def main():
   with open(os.path.join(OUTDIR, "result.svg"),"w") as f:
       f.writelines(res_svg)
 
+  # res_svg2 should match res_svg
+  cfg = img2num.ImageToSvgConfig(km = {"k": 16})
+  print(cfg)
+
+  res_svg2 = img2num.image_to_svg(img, config=cfg)
```
Krasner, line 37: @Ryan-Millard 
calling `img2num.image_to_svg(img)` will use a default config struct. but otherwise `config` becomes a keyword argument so the user has to do `img2num.image_to_svg(img, config=cfg)` as above. ok?

Krasner:
@Ryan-Millard feel free to update the doxygen stuff too...

Krasner:
also I think it would be better to use the step by step pipeline in the react demo rather than the unified call to get the hedgehog to move... unless you got a way to advance the hedgehog while the c++ backend is running

Ryan-Millard:
> also I think it would be better to use the step by step pipeline in the react demo rather than the unified call to get the hedgehog to move... unless you got a way to advance the hedgehog while the c++ backend is running

You are right about that. It is doable, but it would be very bad to do, so let's not do it.

Just run this from the repository's root to undo your changes to it:

```js
git checkout main -- example-apps/react-js/src/components/WasmImageProcessor.jsx
```

It'll checkout to the main branch's current state, but the `--` specifies that it should only be applied to those specific files.

---

What you've done with `image_to_svg` is great, but the config struct has some weird names. Please name them explicitly. For example, `km` should be `kmeans` or `bf` should be `bilateral_filter`/`bilat_filter`. These changes will make it more user-friendly.

Basically, we should aim to keep the API the same across all languages so that we (contributors & users) don't need to memorize how to use it for each language.

Ryan-Millard:
> @Ryan-Millard feel free to update the doxygen stuff too...

Alright. I'll work on it tomorrow. I didn't want to do it because I wasn't sure how you felt about me committing to your branches - it sometimes felt like I was taking over and I didn't want that.

Krasner:
> > @Ryan-Millard feel free to update the doxygen stuff too...
> 
> Alright. I'll work on it tomorrow. I didn't want to do it because I wasn't sure how you felt about me committing to your branches - it sometimes felt like I was taking over and I didn't want that.

I'm more than happy to let someone else do documentation 😅

bindings/js/src/wasm_wrapper.c
Krasner: @Ryan-Millard do you want to fix this in this PR? Changing to `wasm_wrapper.c` and updating the bindings/js CMakeLists.txt is pretty easy and it works on my end.

Krasner: doing it

Ryan-Millard:
Hi @Krasner. Please rather prefer separate PRs for each task you do.

For example, the transition from C++ to C-only for `bindings/js` should not be part of unifying the library into a single function because they serve different purposes - one is a new feature and the other is an internal refactor.

This will become very important when we have our first release (which I plan to have once this PR is merged) because a commit like this will break the Conventional Commit and SemVer standards that the people who use our library will expect us to follow. Also, if you introduced a dangerous bug in the `bindings/js` refactor, we'd have to revert this entire PR because it'll be squashed into one commit - that'll remove the `image_to_svg` function as well.

Keeping it smaller makes it easier to review and spot bugs, too.

Moving foward, please try to keep your pull requests simple - especially when it comes to things that will affect users.

Ryan-Millard:
> > > @Ryan-Millard feel free to update the doxygen stuff too...
> > 
> > 
> > Alright. I'll work on it tomorrow. I didn't want to do it because I wasn't sure how you felt about me committing to your branches - it sometimes felt like I was taking over and I didn't want that.
> 
> I'm more than happy to let someone else do documentation 😅

Nobody likes documentation - that's why we basically have none. The stuff we have is pretty much all outdated.

bindings/py/src/img2num_pybind.cpp
```diff
@@ -120,4 +120,88 @@ PYBIND11_MODULE(_img2num, m) {
...
+                // 4. Process remaining top-level kwargs (like color_space or min_cluster_area)
+                if (kwargs.contains("min_cluster_area"))
+                    c->min_cluster_area = kwargs["min_cluster_area"].cast<int>();
+                if (kwargs.contains("color_space"))
+                    c->color_space = kwargs["color_space"].cast<uint8_t>();
+
+                return c;
+            }), pybind11::arg("bf") = pybind11::dict(), // Defaults to empty dict
+                pybind11::arg("km") = pybind11::dict()  // Defaults to empty dict
+            )
```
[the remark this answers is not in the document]
Ryan-Millard: @coderabbitai you'll have to update this suggestion - it is now outdated. Please resolve this and re-run your review.

Krasner:
@Ryan-Millard what's the status here?

Ryan-Millard (dismissed):

Ryan-Millard:
> @Ryan-Millard what's the status here?

Sorry for the late reply. I was away over the weekend. 

Everything looks good. I'm just going to clean up a few things, then I'll merge it.

Ryan-Millard:
@Krasner, my latest commit clearly separates our logic:

`bindings/py`: simple bindings to make the C++ compatible with Python
`packages/py`: more advanced code to make the API **"Pythonic"**. This includes things like defaults and any special logic that doesn't need to be done at the Pybind stage. 

This separation of concerns protects us from overly scary-looking bindings (C++ and Pybind11 are very verbose and unreadable for the untrained eye, whereas Python is borderline pseudocode).

Krasner:
> @Krasner, my latest commit clearly separates our logic:
> 
> `bindings/py`: simple bindings to make the C++ compatible with Python `packages/py`: more advanced code to make the API **"Pythonic"**. This includes things like defaults and any special logic that doesn't need to be done at the Pybind stage.
> 
> This separation of concerns protects us from overly scary-looking bindings (C++ and Pybind11 are very verbose and unreadable for the untrained eye, whereas Python is borderline pseudocode).

Oh yeah that's a much better solution :)

Ryan-Millard:
@coderabbitai please make a few issues here:

1. Documentation for the changes added in this pull request (be granular and make sub-issues)
2. Cleanup for the Python API around the `ImageToSvgConfig` struct - it needs a more Pythonic API because it may appear a bit weird to normal Python users. It possibly needs a special Python class or a complete rework.

DO NOT BE LAZY @coderabbitai!!!!

Ryan-Millard:
@Krasner please also see #342. Also, please let me know if you are interested in that part of maintaining in this project?

Ryan-Millard:
> @Krasner please also see #342. Also, please let me know if you are interested in that part of maintaining in this project?

@Krasner, are you interested in such parts of the maintenance of Img2Num?

PR #336, opened 2026-04-29
Potential fix for code scanning alert no. 81: Uncontrolled data used in path expression

Ryan-Millard:
Potential fix for [https://github.com/Ryan-Millard/Img2Num/security/code-scanning/81](https://github.com/Ryan-Millard/Img2Num/security/code-scanning/81)

Best fix: validate and constrain CLI-derived paths before file operations. For this script, the safest low-impact approach is to resolve paths and enforce that:
1. `SHADER_DIR` exists and is a directory.
2. `OUTPUT_FILE` resolves to a location inside the current working tree (or another explicit safe root), preventing arbitrary writes outside project scope.

In `core/tools/embed_shaders.py`, replace direct `Path(sys.argv[i])` assignments with resolved paths plus checks. Add a small helper `require_within_root(path, root, label)` that uses `Path.resolve()` and `relative_to()` to enforce containment. Then use validated `OUTPUT_FILE` for reads/writes as before. This preserves existing functionality for expected in-repo usage while preventing traversal/absolute-path abuse.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/core/tools/embed_shaders.py
+++ b/core/tools/embed_shaders.py
@@ -1,7 +1,20 @@
 import sys, pathlib
 
-SHADER_DIR = pathlib.Path(sys.argv[1])
-OUTPUT_FILE = pathlib.Path(sys.argv[2])
+def require_within_root(path: pathlib.Path, root: pathlib.Path, label: str) -> pathlib.Path:
+    resolved = path.resolve()
+    root_resolved = root.resolve()
+    try:
+        resolved.relative_to(root_resolved)
+    except ValueError:
+        raise ValueError(f"{label} must be within {root_resolved}: {resolved}")
+    return resolved
+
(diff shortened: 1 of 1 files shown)


Ryan-Millard:
@Krasner, do you think this is the right solution or do you think we should just hard-code the path instead of taking it as an argument?

Krasner:
> @Krasner, do you think this is the right solution or do you think we should just hard-code the path instead of taking it as an argument?

Let's just hard code them! 😅

Ryan-Millard:
> > @Krasner, do you think this is the right solution or do you think we should just hard-code the path instead of taking it as an argument?
> 
> Let's just hard code them! 😅

Thank you. I'll close this PR now and implement the fix.

PR #341, opened 2026-05-02
fix(docker): copy stb submodule into dev image to fix Python bindings

Geff115:
## Summary

The dev Docker image was missing the `stb` submodule, causing the Python
bindings build to fail with:

fatal error: stb/stb_image.h: No such file or directory

This meant that running `uv sync --no-build-isolation` inside the container
would always error out, blocking any contributor from setting up the Python
bindings locally.

## Root Cause

`Dockerfile.dev` copied `third_party/dawn` into the image, but never copied
`third_party/stb`. Since the Python bindings built via CMake depend on STB
for image loading, the build failed at compile time.

Note: Running `git submodule update --init --recursive` is not a viable
workaround because Dawn's nested sub-submodules include a Google-internal
repository (`chrome-internal.googlesource.com`) that requires special
credentials inaccessible to external contributors.

## Fix

Added `COPY third_party/stb third_party/stb` to `Dockerfile.dev`, alongside
the existing Dawn copy. Contributors must have the `stb` submodule initialized
on their host machine before building the image:

git submodule update --init third_party/stb

## Testing

Verified the full setup sequence inside the container succeeds:

uv venv
uv pip install scikit_build_core numpy
uv sync --no-build-isolation
uv run python3 -c "import img2num;"

All steps completed without errors.

## Related

Raised by @Ryan-Millard in issue #332 thread. This should close the #340 issue created by @coderabbitai in the thread.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -59,6 +59,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 
 # After installing deps since Dawn is likely to change, not its deps
 COPY third_party/dawn third_party/dawn
+COPY third_party/stb third_party/stb
 
 RUN cmake -S third_party/dawn -B build/dawn \
     -DDAWN_SUPPORTS_GLFW_FOR_WINDOWING=OFF \


Ryan-Millard:
Hi. 

Please will you explain more about the problem you faced here.

STB was intentionally not included in the image because of the type of library it is (a header-only C++ library). Dawn, however, was included and built into the image because it takes about 30-45 minutes on a normal computer (the image and repository take long enough to get locally).

If you weren't able to use STB, I'd like to first verify that it wasn't an accident that occurred when setting up your environment (maybe due to outdated documentation). The Docker container mounts to the project’s directory, so it should've just worked without any problems
problems. My suspicion is a submodule-related problem because they behave weirdly sometimes - when you pulled, did you ensure that it recursed the submodules?

Geff115:
Hi @Ryan-Millard, thank you for the clarification!

You're right! It was a submodule initialization issue. When I cloned the repo, I didn't initialize the submodules, so `third_party/stb` was an empty directory on my host machine. Since the container mounts the project directory, it saw the empty folder too, which caused the missing `stb_image.h` error during the build.
Running `git submodule update --init third_party/stb` on the host resolved it immediately.

So rather than a Dockerfile change, the fix should be a documentation update, adding a note to the setup/contributing docs to remind contributors to initialize submodules after cloning. Would you like me to update the PR to reflect that instead?

Ryan-Millard:
> Hi @Ryan-Millard, thank you for the clarification!
> 
> You're right! It was a submodule initialization issue. When I cloned the repo, I didn't initialize the submodules, so `third_party/stb` was an empty directory on my host machine. Since the container mounts the project directory, it saw the empty folder too, which caused the missing `stb_image.h` error during the build. Running `git submodule update --init third_party/stb` on the host resolved it immediately.
> 
> So rather than a Dockerfile change, the fix should be a documentation update, adding a note to the setup/contributing docs to remind contributors to initialize submodules after cloning. Would you like me to update the PR to reflect that instead?

Sure, I'd love that. Thank you so much.

I'd prefer it if you'd rather open up another PR, though. Closing this one adds traceability and makes things clean and easy to look back at in the future (for example, cross-referencing this PR in a future conversation).

I'm going to close this PR since it doesn't fix a problem in the repository. Please will you reference it when you make the new one.

Have a great evening further!🦔🦔🦔

Ryan-Millard:
@coderabbitai, please create an issue for this. Look at the conversation and see that the problem related to this PR was a documentation issue and that it needs to be fixed. Please make sure to link it when the new PR is created to fix the documentation on the [Setup & Dependencies](https://ryan-millard.github.io/Img2Num/info/docs/next/contributing/setup-and-dependencies/) ([`docs/docs/contributing/setup-and-dependencies/index.md`](https://github.com/Ryan-Millard/Img2Num/blob/main/docs/docs/contributing/setup-and-dependencies/index.md)) page for contributors.

Additionally, I noticed that I didn't update all the admonitions with Docusaurus's latest version and it looks like they are now broken on all of the pages (specifically when the admonition has a title). Please create a separate issue for this and do not reference this PR in that issue. You may reference the PR that updates the Docusaurus version (#327) and the [Setup & Dependencies](https://ryan-millard.github.io/Img2Num/info/docs/next/contributing/setup-and-dependencies/) page as an example of the problem (specifically the "Docker Dev Environment" admonition on [line 25](https://github.com/Ryan-Millard/Img2Num/blob/fce52c6779013ee13717f31000b431eb67a2ec42/docs/docs/contributing/setup-and-dependencies/index.md?plain=1#L25)).

PR #346, opened 2026-05-02
fix(packages/js): Remote property injection in wasmWorker.js

Ryan-Millard:
## What was changed & why

Simple codeql vulnerability fix. It was unlikely.

Fixes: #none

## Changes

wasmWorker.js:191 => Object.create(null) for safety

## Testing & Verification

React example works fine


Ryan-Millard:
@coderabbit full review

PR #404, opened 2026-05-28
fix: syntax errors in release.yml and pyproject.toml

Ryan-Millard:
## What was changed & why

Fixes: #<issue-number>


Ryan-Millard:
@coderabbit generate a commit message for this

PR #423, opened 2026-06-01
fix(docker): fix dev user permissions in Dockerfile.dev

Ryan-Millard:
## What was changed & why

`dev` user permissions fixed to be permissive for developers to do what they need to do?

Fixes: none

## Changes

`Dockerfile.dev`
- `dev` user management improved to allow them to access the resources they need to.

PR #429, opened 2026-06-08
fix(core)!: prevent holes during SVG generation

Krasner:
## What was changed & why

Current method produces holes when bezier simplification / smoothing is applied since shared contour edges (those between neighboring regions) are processed independently. Here we treat shared edges correctly so that contour simplification does not cause holes.

Work done using Claude Code - trying many ideas to improve upon current implementation.

Let me know if the results look cleaner and better - from my local tests they seem a bit more pleasing.

Fixes: #<issue-number> 

## Additional Resources

OLD OUTPUT SVG:
cow_old

NEW OUPUT SVG:
cow_new

note the jagged edges in the old svg and holes between edges that should be overlapping

## Summary by CodeRabbit

* **New Features**
  * Added `min_thickness` parameter to SVG generation APIs (C, Python, JavaScript, C++) for filtering thin regions
  * Introduced junction detection and preservation—junction points now remain as exact curve endpoints during fitting
  * Enhanced contour construction pipeline with improved shared-edge handling and smoothing near junctions

* **Chores**
  * Updated Docker configuration and example applications


Ryan-Millard:
Hi. This sounds good, but I still haven't gotten my PC properly fixed, so I can't review it easily.🥲

I'll try my best when I get the chance. Thank you!

Krasner:
> Hi. This sounds good, but I still haven't gotten my PC properly fixed, so I can't review it easily.🥲
> 
> I'll try my best when I get the chance. Thank you!

Oh right, I forgot... Hope you get your new PC soon!

Krasner:
@Ryan-Millard not sure what the wheel building failure in the python ci build is caused by. locally I'm able to do `uv build` in the docker container and build the wheel

Ryan-Millard:
@Krasner, I'm not sure. I think it was because you set the PR to target `main` instead of `dev`, which probably made it run using the workflows on `main` instead. The workflows on `main` were broken at a point in time and I'm not sure if `dev` even has the fixed versions.

Krasner:
> @Krasner, I'm not sure. I think it was because you set the PR to target `main` instead of `dev`, which probably made it run using the workflows on `main` instead. The workflows on `main` were broken at a point in time and I'm not sure if `dev` even has the fixed versions.

correct. `dev` has the updated Dockerfile and associated image that don't match `main`. Let me resolve the conflicts and then try against the `dev` branch...

Ryan-Millard:
> > @Krasner, I'm not sure. I think it was because you set the PR to target `main` instead of `dev`, which probably made it run using the workflows on `main` instead. The workflows on `main` were broken at a point in time and I'm not sure if `dev` even has the fixed versions.
> 
> correct. `dev` has the updated Dockerfile and associated image that don't match `main`. Let me resolve the conflicts and then try against the `dev` branch...

It'll still fail because I broke the image the other day.😂😂

I'm still trying to fix the permissions errors in #423 - it just takes so long to build the stuff.

Dockerfile.dev
```diff
@@ -81,6 +81,7 @@ ENV EMSDK=$EMSDK_DIR
 # Auto-load emsdk in all shells
 RUN echo "source /opt/emsdk/emsdk_env.sh" >> /etc/bash.bashrc
 
+RUN chown -R dev:dev /opt/emsdk/
```
Krasner: @Ryan-Millard this is the missing permission to get the wasm/js build working

Ryan-Millard: @Krasner it's been a while since I tried that fix, but I'm pretty sure that there were more problems inside the container once it was built. I can't fully remember and I think I'm going to die if this is the fix because I put a lot of work into #423 and am still struggling with it.

Krasner: You can see all the ci build tests passed after this change

Ryan-Millard: It's very strange. uv isn't present when I pull the image, yet it works fine in CI.

```sh
dev@bb96aaa15dd8:/usr/src/app$ uv --version
bash: uv: command not found
dev@bb96aaa15dd8:/usr/src/app$ pnpm --version
10.29.1
dev@bb96aaa15dd8:/usr/src/app$ cmake --version
cmake version 3.31.6

CMake suite maintained and supported by Kitware (kitware.com/cmake).
dev@bb96aaa15dd8:/usr/src/app$ echo "$IMG2NUM_IMAGE"
ghcr.io/ryan-millard/img2num-dev:pr-429
dev@bb96aaa15dd8:/usr/src/app$
```

Krasner: @Ryan-Millard check the image again (I regenerated it from an updated Dockerfile). The issue is again the root vs dev permissions. dev did not have access to uv. I also updated docker-compose.yml so pull that too

Krasner:
@Ryan-Millard everything seems to be working.
A few conclusions:
- having a dev user is good for security but makes permissions difficult to manage
- since /usr/src/app mounts to the users host directory, dev needs correct permissions to be allowed to write (this is default for root). For this I added `user:` field to docker-compose
- uv is installed under root by default. I updated Dockerfile.dev to give dev access to uv
- not sure why the github actions behavior sometimes differs from local behavior - i.e when the user pulls the container and runs it but that's important to always check

Ryan-Millard:
> @Ryan-Millard everything seems to be working. A few conclusions:
> 
> * having a dev user is good for security but makes permissions difficult to manage
> * since /usr/src/app mounts to the users host directory, dev needs correct permissions to be allowed to write (this is default for root). For this I added `user:` field to docker-compose
> * uv is installed under root by default. I updated Dockerfile.dev to give dev access to uv
> * not sure why the github actions behavior sometimes differs from local behavior - i.e when the user pulls the container and runs it but that's important to always check

@Krasner please excuse while I die quietly.😢

This means that #423 can be closed.

I just ran into this:
```sh
[00:19] ~/projects/Img2Num $ ./img2num sh --ghcr-img pr-429
WARN[0000] The "UID" variable is not set. Defaulting to a blank string.
WARN[0000] The "GID" variable is not set. Defaulting to a blank string.
[+] up 16/16
 ✔ Image ghcr.io/ryan-millard/img2num-dev:pr-429 Pulled                                                             673.2s
 ✔ Container img2num-dev-1                       Recreated                                                          5.8s
```

...so we just need this in `docker-compose.yml`:
```yml
user: "${UID:-1000}:${GID:-1000}"
```

Krasner:
> > @Ryan-Millard everything seems to be working. A few conclusions:
> > 
> > * having a dev user is good for security but makes permissions difficult to manage
> > * since /usr/src/app mounts to the users host directory, dev needs correct permissions to be allowed to write (this is default for root). For this I added `user:` field to docker-compose
> > * uv is installed under root by default. I updated Dockerfile.dev to give dev access to uv
> > * not sure why the github actions behavior sometimes differs from local behavior - i.e when the user pulls the container and runs it but that's important to always check
> 
> @Krasner please excuse while I die quietly.😢
> 
> This means that #423 can be closed.
> 
> I just ran into this:
> 
> ```shell
> [00:19] ~/projects/Img2Num $ ./img2num sh --ghcr-img pr-429
> WARN[0000] The "UID" variable is not set. Defaulting to a blank string.
> WARN[0000] The "GID" variable is not set. Defaulting to a blank string.
> [+] up 16/16
>  ✔ Image ghcr.io/ryan-millard/img2num-dev:pr-429 Pulled                                                             673.2s
>  ✔ Container img2num-dev-1                       Recreated                                                          5.8s
> ```
> 
> ...so we just need this in `docker-compose.yml`:
> 
> ```yaml
> user: "${UID:-1000}:${GID:-1000}"
> ```

Yes will add the defaults

Ryan-Millard:
@Krasner please will you explain more about this PR and what it does. The output looks very similar to what we already have, so I think I'm not looking at the right things.


**Input image:**
dreamy-african-savannah


**Output image:**
result

Krasner:
Yup it should look similar but better in that: 
1) it guarantees no holes in the SVG - the existing solution has independent bezier fitting of contours which can cause neighboring edges to no longer overlap.
2) it adds a min_thickness argument to label_to_svg - this removes very thin regions and merges them to a neighbor, so it looks visually a little better

Krasner:
> @Krasner please excuse while I die quietly.😢

don't die yet... without the defaults UID and GID are interpreted as 0 (`0:0`) which is actually root. So this makes docker-compose force the docker container to use the `root` user rather than the `dev` user. That's why everything works... 
Maybe it's easier to just remove the `dev` user entirely

To add - in order for `dev` to be able to write to the mounted volume it's user id and group id MUST match the user id and group id on the host system. We can hard code them both to 1000 and for most users this probably will be fine. But in the rare case where the users pulls the docker image but their user id and group id are not 1000, they will have a permission error

Krasner:
@Ryan-Millard final conclusion:
The docker container has to be in root user mode. This allows compatibility with github runners and local machines. Like I mentioned previously, if the docker container is **not** using root user, then the non-root user's user and group ids must match the host systems user and group ids in order to have write access on a mounted volume. For most users it would be 1000, **but** that does not seem to be the case for a github runner - thus all of the ci builds fail.

By setting the docker user to root, this works out of the box on any system. In the latest dockerfile change, I remove the `dev` user entirely.

Ryan-Millard:
> * uv is installed under root by default. I updated Dockerfile.dev to give dev access to uv
> * not sure why the github actions behavior sometimes differs from local behavior - i.e when the user pulls the contai

Do you think it's even worthwhile to keep the `dev` user? I've been trying to find a clean way to make it work and it just feels like something that will grow in complexity over time and introduces issues that we don't really want to be focusing on. The security risk does exist, but I feel like it would be tolerable because it's just a development environment.

Krasner:
> > * uv is installed under root by default. I updated Dockerfile.dev to give dev access to uv
> > * not sure why the github actions behavior sometimes differs from local behavior - i.e when the user pulls the contai
> 
> Do you think it's even worthwhile to keep the `dev` user? I've been trying to find a clean way to make it work and it just feels like something that will grow in complexity over time and introduces issues that we don't really want to be focusing on. The security risk does exist, but I feel like it would be tolerable because it's just a development environment.

Let's just remove the dev user altogether. That's what I've done in the last Dockerfile update. It's the most robust solution

Ryan-Millard:
> > > * uv is installed under root by default. I updated Dockerfile.dev to give dev access to uv
> > > * not sure why the github actions behavior sometimes differs from local behavior - i.e when the user pulls the contai
> > 
> > 
> > Do you think it's even worthwhile to keep the `dev` user? I've been trying to find a clean way to make it work and it just feels like something that will grow in complexity over time and introduces issues that we don't really want to be focusing on. The security risk does exist, but I feel like it would be tolerable because it's just a development environment.
> 
> Let's just remove the dev user altogether. That's what I've done in the last Dockerfile update. It's the most robust solution

That's alright, then. I will still be annoyed when working locally because I don't have root access locally on my machine, but whatever.😢

---

Please will you show me some comparisons between the old and new outputs for this PR so I can see the differences more clearly.

Krasner:
> > > > * uv is installed under root by default. I updated Dockerfile.dev to give dev access to uv
> > > > * not sure why the github actions behavior sometimes differs from local behavior - i.e when the user pulls the contai
> > > 
> > > 
> > > Do you think it's even worthwhile to keep the `dev` user? I've been trying to find a clean way to make it work and it just feels like something that will grow in complexity over time and introduces issues that we don't really want to be focusing on. The security risk does exist, but I feel like it would be tolerable because it's just a development environment.
> > 
> > 
> > Let's just remove the dev user altogether. That's what I've done in the last Dockerfile update. It's the most robust solution
> 
> That's alright, then. I will still be annoyed when working locally because I don't have root access locally on my machine, but whatever.😢
> 
> Please will you show me some comparisons between the old and new outputs for this PR so I can see the differences more clearly.

OLD SVG:
cow_old

NEW SVG:
cow_new

note the jagged edges in the old svg and holes between edges that should be overlapping

Krasner:
@Ryan-Millard this is ready for merging.

Krasner:
@Ryan-Millard have you had time to review?

Ryan-Millard:
> @Ryan-Millard have you had time to review?

Unfortunately not. WSL stopped working on my laptop yesterday and I've been trying to get it back up since then. I can't run Docker on the laptop either because it uses WSL. I think there's something wrong with the hardware.

The new CPU is coming next week, so I'll be able to use my PC when it arrives.

---

Other than that, I'm a bit concerned about how to handle this kind of situation in the future. 

**1. We don't want to constantly break the API with every new feature (this one would force existing users to change how they use the library).** We probably need to discuss how to manage this kind of thing because I can imagine it becoming quite annoying for our users if we keep changing the API and break things for then when they update.

**2. Would you be willing to take care of the repository for a while now (until my PC is working) and potentially in the future (if I run into something similar again)?** This would include:

- Reviewing PRs from others
- Keeping the issue log neat and relevant to the current state of the repository
- Guiding newcomers
- General maintenance

Krasner:
@Ryan-Millard what's the process of merging dev to main? When do we determine that there are enough changes in dev that it's time to merge to main?

Ryan-Millard:
> @Ryan-Millard what's the process of merging dev to main? When do we determine that there are enough changes in dev that it's time to merge to main?

I will `git checkout main && git merge dev --ff-only && git push` when we know that `dev` is stable. We only really need to update it when the documentation and library changes. Fast forward-only merges preserve the history and is the least aggressive way to sync the two branches.

Currently, only I can push to main because GitHub is a bit daft when it comes to permissions on personal repositories - the roles are flat, so adding a new collaborator gives them the permissions you have (that's what scares me). I've been thinking about moving to an org because GitHub orges have better role structures, but it'll be a big change, so I haven't done that.

For now, it'll just be me managing `main` so that we don't run into accidents and potentially dangerous situations when we get new maintainers in the future.

Krasner:
@Ryan-Millard please review this PR

Ryan-Millard:
> @Ryan-Millard please review this PR

Hi. I'm looking at it now.

This is a breaking change since the API has changed and people using the library will need to change how the use it. We follow the [Conventional Commit](https://gist.github.com/Zekfad/f51cb06ac76e2457f11c80ed705c95a3) style to allow `release-please` to automatically generate docs for us.

As a result, the commit for this will follow `fix(...)!`.

core/include/internal/bezier.h
```diff
@@ -3,8 +3,14 @@
...
+void fit_curve_reduction(const std::vector<std::vector<Point>> &chains,
+                         std::vector<std::vector<QuadBezier>> &results, float tolerance);
+
+// Same, but `fixed[i][k]!=0` marks point k of chain i as a junction that must NOT
+// move: the chain is split at those points so each becomes an exact (pinned)
+// curve endpoint. Chains with no fixed points fit identically to the overload
+// above.
+void fit_curve_reduction(const std::vector<std::vector<Point>> &chains,
+                         const std::vector<std::vector<uint8_t>> &fixed,
+                         std::vector<std::vector<QuadBezier>> &results, float tolerance);
```
Ryan-Millard: Please explain what is going in the comment here. Should it be updated?

Additionally, do you use both in this code?

core/include/internal/douglas_peucker.h
```diff
@@ -0,0 +1,28 @@
...
+// that share a junction therefore split at the SAME point. Simplification is
+// bounded so it never retracts a boundary inward past the inter-region overlap
+// margin, so reducing one region never eats its overlap with another and opens a
+// gap. Kept points are emitted as straight-line quads, so the existing QuadBezier
+// SVG path emission (and its closing 'Z') is unchanged.
+//
+// `eps`        : overall deviation tolerance (px). Env override IMG2NUM_DP_EPS.
+// retract_eps  : max inward (region-shrinking) move allowed when dropping a
+//                vertex; defaults to min(eps, 0.5px). Env override
+//                IMG2NUM_DP_RETRACT.
```
Ryan-Millard: Please make this a Doxygen doc string - it's already basically there. CodeRabbit already suggested a fairly decent one, so go check [that one](https://github.com/Ryan-Millard/Img2Num/pull/429/changes#r3374322387).

Also, please try to mention technical stuff in your future PRs so I know to expect this kind of thing - I completely forgot about this algorithm and that I was thinking about it a while ago. Thank you for implementing it.

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     pull_policy: always
     working_dir: /usr/src/app
     volumes:
-      - .:/usr/src/app:cached
+      - .:/usr/src/app:rw,cached
```
Ryan-Millard: Accident?

Ryan-Millard (changes requested):
Hi @Krasner. This is great stuff and now that I see it's the Douglas-Peucker implementation, I understand exactly what to look for. I can also see the differences more easily. Thank you for implementing this!

Before it gets merged, please will you update the documentation (specifically the stuff for user-facing code) because that is very important (users need to know how things work). CodeRabbit already has a few suggestions, so please will you go through everything CodeRabbit suggested and apply the stuff where necessary. 

Once again, this is great stuff. Thank you!

Ryan-Millard:
@coderabbitai please generate a good commit message as well as a description / patch notes for this. I will be squashing this PR, so I need notes that work for everything here.

core/include/internal/bezier.h
Krasner: Yes, the junction points are critical, i.e where 3 or more neighbors touch. Those points must not move - they are fixed. In the curve simplification/smoothing step those points must marked as unmoveable otherwise there will be a hole.

docker-compose.yml
Krasner: I can remove it. it think `rw` is by default.

Ryan-Millard (dismissed):

core/src/internal/bezier.cpp
```diff
@@ -150,3 +150,37 @@ void fit_curve_reduction(const std::vector<std::vector<Point>> &chains,
...
+        // Fit each [bounds[s], bounds[s+1]] piece; consecutive pieces share the
+        // junction point, so the curve stays continuous and pinned there.
+        for (size_t s = 0; s + 1 < bounds.size(); ++s) {
+            const int a = bounds[s], b = bounds[s + 1];
+            std::vector<Point> seg(chain.begin() + a, chain.begin() + b + 1);
+            fitRecursive(seg, tolerance, result);
+        }
+        results.push_back(result);
+    }
+}
```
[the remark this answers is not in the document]
Ryan-Millard: @Krasner please check this out. I'm not sure if this is even likely.

core/src/internal/douglas_peucker.cpp
```diff
@@ -0,0 +1,110 @@
...
+            if (prev >= 0) {
+                const Point &P = chain[prev];
+                const Point &Q = chain[k];
+                result.push_back({P, (P + Q) * 0.5f, Q});
+            }
+            prev = k;
+        }
+        results.push_back(result);
+    }
+}
```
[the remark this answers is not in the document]
Ryan-Millard: @Krasner please check this out.

I also noticed that `IMG2NUM_DP_EPS` doesn't seem to be defined anywhere. Shouldn't they rather be arguments to the function?

core/src/internal/graph.cpp
```diff
@@ -205,44 +209,177 @@ void Graph::process_overlapping_edges() {
...
+            for (std::vector<QuadBezier> &curve : it->second) {
+                std::vector<Point> anchors;  // keep contours[] parallel to curves[]
+                anchors.reserve(curve.size() + 1);
+                for (const QuadBezier &q : curve) anchors.push_back(q.p0);
+                if (!curve.empty()) anchors.push_back(curve.back().p2);
+                n->m_contours.contours.push_back(std::move(anchors));
+                n->m_contours.curves.push_back(std::move(curve));
+                n->m_contours.colors.push_back(col);
+                n->m_contours.hierarchy.push_back({-1, -1, -1, -1});
+                n->m_contours.is_hole.push_back(false);
```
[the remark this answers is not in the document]
Ryan-Millard: @Krasner please check this out. I understand that we don't really use the hierarchical information because we're more focused on the actual edges, but we may want to preserve this in the future.

core/src/internal/shared_contours.cpp
```diff
@@ -0,0 +1,293 @@
...
+#include <set>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "internal/bezier.h"
+
+namespace {
+
+constexpr int32_t OUTSIDE = -2147483647;  // image exterior label
```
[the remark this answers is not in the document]
Ryan-Millard: @Krasner, I also think it's better to use the minimum value constants provided by the standard library.

bindings/py/src/img2num_pybind.cpp
```diff
@@ -113,17 +113,17 @@ PYBIND11_MODULE(_img2num, m) {
...
 
-            std::string svg{img2num::labels_to_svg(data_ptr, labels_ptr, width, height, min_area)};
+            std::string svg{img2num::labels_to_svg(data_ptr, labels_ptr, width, height, min_area, min_thickness)};
             pybind11::str svg_py_str(std::move(svg));
 
             return svg_py_str;
         },
         pybind11::arg("data"), pybind11::arg("labels"), pybind11::arg("width"),
-        pybind11::arg("height"), pybind11::arg("min_area"), "Convert labels to SVG string");
+        pybind11::arg("height"), pybind11::arg("min_area"), pybind11::arg("min_thickness"), "Convert labels to SVG string");
```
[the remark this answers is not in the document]
Ryan-Millard: @Krasner, I thought we already had this done. I'm sorry for missing - this is very important.

Krasner:
@Ryan-Millard feel free to merge when you're ready

core/src/internal/graph.cpp
```diff
@@ -205,86 +209,314 @@ void Graph::process_overlapping_edges() {
...
+        /* This is the new method
+        Shared-edge mode (default): build region boundaries on the crack grid so
+        neighbouring contours are exactly coincident along shared edges -- no
+        overlap band, no gaps
+        */
 
-    int j = 0;
-    for (const Node_ptr &n : get_nodes()) {
-        if (n->area() == 0) continue;
+        float eps = 0.25f;
```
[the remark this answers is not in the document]
Ryan-Millard: @Krasner is this meant to represent $\epsilon$ (epsilon)?

Additionally, why is this passed through multiple functions all the way down to `fit_curve_reduction` (as `tolerance`), then over to `fitRecursive` (as `errorLimit`)?

```mermaid
flowchart TD
  A["Graph::compute_contours<br /><sub>float eps = 0.25f;</sub>"] --> B[build_shared_loops]
  B --> C["fit_edge"]
  C --> D["fit_curve_reduction"]
  D --> E["fitRecursive"]
```

It's a hard-coded value that just gets delegated down the call stack. In React, this would be called prop-drilling and it's generally a sign of vulnerable code - I'm not saying that there's anything wrong with it, I'm just making a suggestion and want to see if you had a particular motivation for this.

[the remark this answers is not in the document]
Krasner: epsilon is meant to control the error of the bezier fit. it does get pushed thru the `build_shared_loops` funciton to `fit_edge` to `fit_curve_reduction`. do you propose a better method?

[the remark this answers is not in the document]
Ryan-Millard: I'm not sure. If it is meant to work that way, it's fine. That entire Graph function is a bit confusing, though because there's a fair bit of dead code:

It contains an [`if (true)`](https://github.com/Krasner/Img2Num/blob/37309f0928943c96b54fe7e298270264e8d04f85/core/src/internal/graph.cpp#L262) as well as an else block - the compiler will just strip the if-check and the else block entirely.

Ryan-Millard:
> @Ryan-Millard feel free to merge when you're ready

I think I need to have a better look at this. I seem to have missed a few things (probably because it's late and I'm a bit distracted).

Krasner:
> > @Ryan-Millard feel free to merge when you're ready
> 
> I think I need to have a better look at this. I seem to have missed a few things (probably because it's late and I'm a bit distracted).

no worries. it's hard to find comments in the diff viewer, but to address one comment in the `Graph::compute_contours` method, I do have a big `if (true)` block. The `else` block is the old method. I can delete it but kept it there just in case...

Ryan-Millard:
> > > @Ryan-Millard feel free to merge when you're ready
> > 
> > 
> > I think I need to have a better look at this. I seem to have missed a few things (probably because it's late and I'm a bit distracted).
> 
> no worries. it's hard to find comments in the diff viewer, but to address one comment in the `Graph::compute_contours` method, I do have a big `if (true)` block. The `else` block is the old method. I can delete it but kept it there just in case...

In case of what? There are also cleaner ways to go about keeping old code you don't have a use for - maybe `git stash` it or create a branch to store it. There has to be another solution and I'm going to look for it now.

Ryan-Millard:
> > > > @Ryan-Millard feel free to merge when you're ready
> > > 
> > > 
> > > I think I need to have a better look at this. I seem to have missed a few things (probably because it's late and I'm a bit distracted).
> > 
> > 
> > no worries. it's hard to find comments in the diff viewer, but to address one comment in the `Graph::compute_contours` method, I do have a big `if (true)` block. The `else` block is the old method. I can delete it but kept it there just in case...
> 
> In case of what? There are also cleaner ways to go about keeping old code you don't have a use for - maybe `git stash` it or create a branch to store it. There has to be another solution and I'm going to look for it now.

Since it's the old version, we can use git to get it back (it would be what's called "restoring the hunk" where the hunk is the block of code that changed). It's very simple to do - you just need to be able to use `git blame`, some diffs and restore/similar.

Krasner:
> > > > @Ryan-Millard feel free to merge when you're ready
> > > 
> > > 
> > > I think I need to have a better look at this. I seem to have missed a few things (probably because it's late and I'm a bit distracted).
> > 
> > 
> > no worries. it's hard to find comments in the diff viewer, but to address one comment in the `Graph::compute_contours` method, I do have a big `if (true)` block. The `else` block is the old method. I can delete it but kept it there just in case...
> 
> In case of what? There are also cleaner ways to go about keeping old code you don't have a use for - maybe `git stash` it or create a branch to store it. There has to be another solution and I'm going to look for it now.

I can think of a few cases:
1. We find that the output is unsatisfactory - though it has looked decent for all the images that I've thrown at it
2. A new developer may want to see previous method / chains of thought / approaches that have been tried in case they want to propose a better method (this is probably the more important case).

Sure the one could scrub through various branches, but this is a core part of the algorithm that have not yet been fully stabilized, so having clearer visibility to different versions might be useful

Ryan-Millard:
> > > > > @Ryan-Millard feel free to merge when you're ready
> > > > 
> > > > 
> > > > I think I need to have a better look at this. I seem to have missed a few things (probably because it's late and I'm a bit distracted).
> > > 
> > > 
> > > no worries. it's hard to find comments in the diff viewer, but to address one comment in the `Graph::compute_contours` method, I do have a big `if (true)` block. The `else` block is the old method. I can delete it but kept it there just in case...
> > 
> > 
> > In case of what? There are also cleaner ways to go about keeping old code you don't have a use for - maybe `git stash` it or create a branch to store it. There has to be another solution and I'm going to look for it now.
> 
> I can think of a few cases:
> 
> 1. We find that the output is unsatisfactory - though it has looked decent for all the images that I've thrown at it
> 2. A new developer may want to see previous method / chains of thought / approaches that have been tried in case they want to propose a better method (this is probably the more important case).
> 
> Sure the one could scrub through various branches, but this is a core part of the algorithm that have not yet been fully stabilized, so having clearer visibility to different versions might be useful

I think suggesting that they run git diffing on the function would be the best solution. Since it's been around for a while and hasn't changed, they wouldn't need to be very accurate between the commits.

Ryan-Millard:
@Krasner please will you also update the `README.md` and maybe the Docusaurus homepage if there are any necessary updates. I think that the `README.md` update may be a good idea.

Krasner:
> @Krasner please will you also update the `README.md` and maybe the Docusaurus homepage if there are any necessary updates. I think that the `README.md` update may be a good idea.

Is there anything there to update?

I'm going to have a big documentation PR ready soon like #388 (which can be closed for now) where I will update as much documentation as I can for all the latest changes and new stuff (python bindings)

Ryan-Millard:
> > @Krasner please will you also update the `README.md` and maybe the Docusaurus homepage if there are any necessary updates. I think that the `README.md` update may be a good idea.
> 
> Is there anything there to update?
> 
> I'm going to have a big documentation PR ready soon like #388 (which can be closed for now) where I will update as much documentation as I can for all the latest changes and new stuff (python bindings)

I was talking about the images. Do you think it would be a good idea to update them?

Krasner:
> I was talking about the images. Do you think it would be a good idea to update them?

Yes good point. I've updated them

bindings/c/src/cimg2num.cpp
```diff
@@ -90,22 +92,19 @@ void img2num_bilateral_filter(
...
             if (!result) {
                 return; // Allocation failed
             }
             std::memcpy(result, svg.c_str(), svg.size() + 1);
         },
-        data, labels, width, height, min_area
-    );
+        data, labels, width, height, min_area, min_thickness);
     return result;
 }
```
Ryan-Millard: What are your thoughts on us removing this function entirely in the future (not now) to make it easier to maintain the library: we only expose one interface for converting to SVGs.

bindings/py/src/img2num_pybind.cpp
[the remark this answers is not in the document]
Ryan-Millard: Why would it fail? Internally, the C++ applies a default value. Look at [`core/include/img2num.h`](https://github.com/Ryan-Millard/Img2Num/blob/main/core/include/img2num.h) - it applies the zero there.

Are you saying that Python will throw an error?

core/include/internal/douglas_peucker.h
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * `@param` chains Input polyline chains
+ * `@param` fixed Junction mask: fixed[i][k] != 0 marks point k of chain i as a junction to preserve
+ * `@param` results Output straight-line QuadBezier segments for each chain
+ * `@param` eps Overall deviation tolerance in pixels (overridable via IMG2NUM_DP_EPS env var)
+ *
+ * `@note` retract_eps (max inward boundary move) defaults to min(eps, 0.5px) and can be
+ *       overridden via IMG2NUM_DP_RETRACT environment variable.
+ */
+void dp_curve_reduction(const std::vector<std::vector<Point>> &chains,
```
Ryan-Millard: Please add a reference for this - I think it would be a good idea to start including links to make it easier for future developers to get started. 

```sh
ryan@Ryans-PC:~/projects/Img2Num$ cat doxygen/Doxyfile.base | grep markdown -i
MARKDOWN_SUPPORT       = YES
# --- Markdown ---
```

We have markdown support, so you can link to it using this style:
```md
[text](link)
```

core/include/internal/graph.h
```diff
@@ -39,11 +39,37 @@ class Graph {
...
+     * `@param` w Image width
+     * `@param` h Image height
+     * `@param` x Pixel x-coordinate
+     * `@param` y Pixel y-coordinate
+     * `@return` Pixel value at (x, y), or 0 if out of bounds
+     */
+    inline uint8_t getPixel(const std::vector<uint8_t>& img, int w, int h, int x, int y) {
+        if (x < 0 || x >= w || y < 0 || y >= h) return 0; // Boundary check
+        return img[y * w + x];
+    }
```
Ryan-Millard, line 56: We should probably schedule a refactor to make the library use the image class wherever possible because we repeat this pattern a lot.

core/src/internal/douglas_peucker.cpp
```diff
@@ -0,0 +1,110 @@
...
+// Standard DP keeps the max-deviation vertex while the chord error exceeds eps.
+// In addition we refuse to drop a run if doing so would pull the boundary
+// *inward* (toward the region interior) by more than retract_eps at any vertex.
+// `interior_sign` maps signed_left_dist to an interior(+)/exterior(-) sign for
+// this polygon's winding, so dropping a convex bulge (exterior vertex) counts as
+// retraction while flattening a concave dent (interior vertex) only grows the
+// region and is always safe. Keeping retraction within the inter-region overlap
+// margin is what prevents this from opening holes between neighbours.
+static void dp_reduce(const std::vector<Point> &pts, int lo, int hi, float eps,
+                      float retract_eps, float interior_sign, std::vector<uint8_t> &keep) {
```
Ryan-Millard:
```suggestion
/**
 * @brief Non-retracting Douglas–Peucker simplification for a polyline segment.
 *
 * Recursively simplifies the polyline defined by pts[lo..hi] using a Douglas–Peucker
 * strategy with an additional geometric constraint to prevent inward boundary collapse.
 *
 * The standard DP algorithm retains the point with maximum perpendicular deviation
 * from the chord (lo, hi) if that deviation exceeds `eps`. This variant introduces an
 * additional "non-retraction" rule: a candidate simplification is rejected if removing
 * intermediate vertices would cause the boundary to move inward (toward the region
 * interior) by more than `retract_eps` at any vertex.
 *
 * The sign of inward/outward movement is determined using `interior_sign`, which maps
 * signed left-distance to an interior (+) / exterior (-) convention based on the polygon
 * winding order. As a result:
 * - Removing convex (exterior) bulges is treated as potential retraction.
 * - Flattening concave (interior) dents is always considered safe, as it expands or
 *   preserves the region.
 *
 * This constraint ensures that simplification does not create excessive inward
 * collapse, which could otherwise introduce gaps between neighboring regions or
 * polygons separated by a limited overlap margin.
 *
 * @param pts            Input polyline points.
 * @param lo             Start index of the segment (inclusive).
 * @param hi             End index of the segment (inclusive).
 * @param eps            Maximum allowed perpendicular deviation for DP simplification.
 * @param retract_eps    Maximum allowed inward boundary retraction per vertex.
 * @param interior_sign  Sign mapping used to classify inward vs outward deviation
 *                       based on polygon winding.
 * @param keep           Output mask indicating which vertices are preserved.
 */
static void dp_reduce(const std::vector<Point> &pts, int lo, int hi, float eps,
                      float retract_eps, float interior_sign, std::vector<uint8_t> &keep) {
```

core/src/internal/douglas_peucker.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    keep[split] = 1;
+    dp_reduce(pts, lo, split, eps, retract_eps, interior_sign, keep);
+    dp_reduce(pts, split, hi, eps, retract_eps, interior_sign, keep);
+}
+
+void dp_curve_reduction(const std::vector<std::vector<Point>> &chains,
+                        const std::vector<std::vector<uint8_t>> &fixed,
+                        std::vector<std::vector<QuadBezier>> &results, float eps) {
+    if (const char *e = std::getenv("IMG2NUM_DP_EPS")) eps = static_cast<float>(std::atof(e));
```
Ryan-Millard: @Krasner, why are you using environment variables like `IMG2NUM_DP_EPS`? Shouldn't the user rather pass it as an argument or something similar?

core/src/internal/douglas_peucker.cpp
Ryan-Millard: _⚠️ Potential issue_ | _🟡 Minor_ | _⚡ Quick win_

**Validate that `fixed` dimensions match `chains` dimensions.**

Similar to `bezier.cpp`, this function assumes `fixed[i].size()` is compatible with `chains[i].size()` but only checks the size defensively at line 89. If a caller provides mismatched dimensions, interior junctions may be silently ignored. Consider adding upfront validation that `fixed.size() == chains.size()` and optionally that each `fixed[i].size() == chains[i].size()` to fail fast on contract violations.





<details>
<summary>🛡️ Proposed validation</summary>

```diff
 void dp_curve_reduction(const std::vector<std::vector<Point>> &chains,
                         const std::vector<std::vector<uint8_t>> &fixed,
                         std::vector<std::vector<QuadBezier>> &results, float eps) {
     if (const char *e = std::getenv("IMG2NUM_DP_EPS")) eps = static_cast<float>(std::atof(e));
     float retract_eps = std::min(eps, 0.5f);
     if (const char *r = std::getenv("IMG2NUM_DP_RETRACT"))
         retract_eps = static_cast<float>(std::atof(r));

+    if (fixed.size() != chains.size()) {
+        throw std::invalid_argument("fixed and chains must have the same outer dimension");
+    }
+
     for (size_t i = 0; i < chains.size(); ++i) {
         const std::vector<Point> &chain = chains[i];
         const int n = static_cast<int>(chain.size());
+        if (fixed[i].size() != static_cast<size_t>(n)) {
+            throw std::invalid_argument("fixed[i] and chains[i] must have the same size");
+        }
```
</details>



<details>
<summary>📝 Committable suggestion</summary>

> ‼️ **IMPORTANT**
> Carefully review the code before committing. Ensure that it accurately replaces the highlighted code, contains no missing lines, and has no issues with indentation. Thoroughly test & benchmark the code to ensure it meets the requirements.

```suggestion
void dp_curve_reduction(const std::vector<std::vector<Point>> &chains,
                        const std::vector<std::vector<uint8_t>> &fixed,
                        std::vector<std::vector<QuadBezier>> &results, float eps) {
    if (const char *e = std::getenv("IMG2NUM_DP_EPS")) eps = static_cast<float>(std::atof(e));
    float retract_eps = std::min(eps, 0.5f);
    if (const char *r = std::getenv("IMG2NUM_DP_RETRACT"))
        retract_eps = static_cast<float>(std::atof(r));

    if (fixed.size() != chains.size()) {
        throw std::invalid_argument("fixed and chains must have the same outer dimension");
    }

    for (size_t i = 0; i < chains.size(); ++i) {
        const std::vector<Point> &chain = chains[i];
        const int n = static_cast<int>(chain.size());
        if (fixed[i].size() != static_cast<size_t>(n)) {
            throw std::invalid_argument("fixed[i] and chains[i] must have the same size");
        }
        std::vector<QuadBezier> result;
        if (n < 2) {
            results.push_back(result);
            continue;
        }

        // Orientation: sign of the shoelace area decides which side of a chord is
        // the region interior (see dp_reduce). Computed over the closed chain.
        double sa = 0.0;
        for (int k = 0; k < n; ++k) {
            const Point &p = chain[k];
            const Point &q = chain[(k + 1) % n];
            sa += static_cast<double>(p.x) * q.y - static_cast<double>(q.x) * p.y;
        }
        const float interior_sign = (sa > 0.0) ? 1.0f : -1.0f;

        // Segment boundaries: chain ends plus every interior fixed point. DP runs
        // inside each [bounds[s], bounds[s+1]] span with its ends pinned.
        std::vector<int> bounds;
        bounds.push_back(0);
        for (int k = 1; k < n - 1; ++k)
            if (k < static_cast<int>(fixed[i].size()) && fixed[i][k]) bounds.push_back(k);
        bounds.push_back(n - 1);

        std::vector<uint8_t> keep(n, 0);
        for (int b : bounds) keep[b] = 1;
        for (size_t s = 0; s + 1 < bounds.size(); ++s)
            dp_reduce(chain, bounds[s], bounds[s + 1], eps, retract_eps, interior_sign, keep);

        // Emit kept points in order as straight-line quads.
        int prev = -1;
        for (int k = 0; k < n; ++k) {
            if (!keep[k]) continue;
            if (prev >= 0) {
                const Point &P = chain[prev];
                const Point &Q = chain[k];
                result.push_back({P, (P + Q) * 0.5f, Q});
            }
            prev = k;
        }
        results.push_back(result);
    }
}
```

</details>



<details>
<summary>🧰 Tools</summary>

<details>
<summary>🪛 Cppcheck (2.20.0)</summary>

[style] 57-57: The function 'dp_curve_reduction' is never used.

(unusedFunction)

</details>

</details>

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@core/src/internal/douglas_peucker.cpp` around lines 57 - 110, The function
dp_curve_reduction assumes fixed has the same outer size as chains and uses
fixed[i].size() later; add upfront validation: check fixed.size() ==
chains.size() and for each i verify fixed[i].size() == chains[i].size() (or at
least >= chains[i].size()) and fail fast (e.g., return/error/throw or
assert/log) if not; perform this validation at the start of dp_curve_reduction
before any processing so mismatched dimensions don't silently drop interior
junctions used by dp_reduce and the bounds/fixed logic.
```

</details>

core/src/internal/graph.cpp
```diff
@@ -214,90 +221,229 @@ void Graph::process_overlapping_edges() {
...
+                junction_map[y * w + x] = 1;
+            }
+
+            /*if (neighbors <= 1) {
+                map[y * w + x] = 2; // Endpoint (or isolated)
+            } else if (transitions >= 3) {
+                map[y * w + x] = 3; // Junction (3+ branches)
+            } else {
+                map[y * w + x] = 1; // Line (Body or Corner)
+            }*/
```
Ryan-Millard:
```suggestion
```

core/src/internal/shared_contours.cpp
Ryan-Millard: If you want this to be the absolute lowest possible number, please use the built-in constants:

```suggestion
#include <vector>
#include <limits>

#include "internal/bezier.h"

namespace {

constexpr int32_t OUTSIDE = std::numeric_limits<int32_t>::min();  // image exterior label
```


---

@coderabbitai also said this, but I don't think it's of much importance:

_⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_

**Don't encode the exterior as an in-band region id.**

`build_shared_loops()` is documented to accept any `int32_t` label as a real region id, but `OUTSIDE` steals `-2147483647` for the canvas exterior. If a caller legitimately uses that id, the exterior is merged into that region and its outer boundary is computed incorrectly. Represent the exterior out-of-band instead of reserving a valid label value.

<details>
<summary>🤖 Prompt for AI Agents</summary>

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In `@core/src/internal/shared_contours.cpp` at line 15, Remove the in-band OUTSIDE
sentinel and instead represent the canvas exterior out-of-band: delete the
OUTSIDE constant in shared_contours.cpp and change build_shared_loops() (and its
callers) to accept an explicit exterior marker (e.g. std::optional<int32_t>
exterior_id or a separate bool/flag) rather than treating a numeric label as
"outside"; update all checks that compared labels to OUTSIDE to use the new
out-of-band indicator so a real region id equal to -2147483647 is no longer
co-opted and the exterior boundary handling is kept separate.
```

</details>





****

core/src/internal/shared_contours.cpp
```diff
@@ -0,0 +1,293 @@
...
+        p.swap(q);
+    }
+}
+
+// Smooth a corner chain (endpoints fixed) and fit it to quadratic beziers. Done
+// once per canonical edge; both adjacent regions reuse the result, so the fitted
+// curve is shared and the two regions stay exactly coincident.
+std::vector<QuadBezier> fit_edge(const std::vector<Point> &corners, int w, int h, float eps) {
+    std::vector<Point> pts = corners;
+    smooth_edge(pts, w, h, 5);
```
Ryan-Millard: The magic number 5 for smoothing iterations should be a named constant (e.g., constexpr int SMOOTHING_ITERATIONS = 5;) or exposed as a parameter if callers need control over smoothing strength.


```suggestion
    constexpr int SMOOTHING_ITERATIONS{5};
    smooth_edge(pts, w, h, SMOOTHING_ITERATIONS);
```

core/src/internal/shared_contours.cpp
```diff
@@ -0,0 +1,293 @@
...
+#include <cmath>
+#include <map>
+#include <set>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "internal/bezier.h"
+
+namespace {
```
Ryan-Millard: What is the purpose of the namespace in this file?

core/src/internal/shared_contours.cpp
```diff
@@ -0,0 +1,293 @@
...
+    for (int y = 0; y < h; ++y)
+        for (int x = 0; x < w; ++x) {
+            int32_t r = L(x, y);
+            auto &dir = region_dir[r];
+            const int c00 = cidx(x, y), c10 = cidx(x + 1, y);
+            const int c11 = cidx(x + 1, y + 1), c01 = cidx(x, y + 1);
+            if (L(x, y - 1) != r) dir[c00].push_back(c10);  // top  ->
+            if (L(x + 1, y) != r) dir[c10].push_back(c11);  // right v
+            if (L(x, y + 1) != r) dir[c11].push_back(c01);  // bottom <-
+            if (L(x - 1, y) != r) dir[c01].push_back(c00);  // left  ^
```
Ryan-Millard: Your comments look wrong - the arrows don't match the words you wrote.

Ryan-Millard (dismissed):
@Krasner this all looks good and it can be merged as-is. Please just address the comments I left as well as run this:

```sh
pnpm format:cpp # there's a bug with the format script
```

bindings/c/src/cimg2num.cpp
Krasner: I'd still keep it... it's not difficult to maintain

core/src/internal/douglas_peucker.cpp
Krasner: removed

PR #433, opened 2026-06-11
fix(packages/js): add node js support via `worker_threads`

Prachi-Gupta2808:
## What was changed & why

The `img2num` JS package used the browser-only Web Workers API, causing a `ReferenceError: Worker is not defined` crash in any Node.js environment. This PR adds Node.js support by detecting the runtime environment and using `worker_threads` instead of Web Workers when running in Node.js.

Fixes: #394

## Changes

- `packages/js/wasmClient.js` : added environment detection; uses `worker_threads` Worker in Node.js, existing Web Worker in browser ( Didn't change anything in this flow)
- `packages/js/wasmWorker.js` : extracted message handler into `handleMessage()`; uses `parentPort` for messaging in Node.js, existing `self.onmessage`/`self.postMessage` in browser (Didn't change anything for browser here either)
- `bindings/js/CMakeLists.txt` : added `node` to `ENVIRONMENT` flag so the compiled WASM supports Node.js runtime
- `example-apps/console-js` : new example app demonstrating the library working in a Node.js environment

## Testing & Verification

Tested inside the Docker container (`main` image as mentioned by @Ryan-Millard ):
1. Reproduced the original `Worker is not defined` error
2. Applied the fix
3. Built the WASM with the updated `ENVIRONMENT` flag
4. Ran `node index.js` in `example-apps/console-js` , library loaded and processed successfully in Node.js

Browser path is completely untouched so existing functionality is unaffected.

## Additional Resources

Note: I already mentioned the webGPU one stuff.

Thank you!

PR #439, opened 2026-06-16
feat: Just command manager

Krasner:
## What was changed & why

Enable Just in docker environment. Create initial justfile.

Fixes: #391


Krasner:
@Ryan-Millard please contribute directly to this branch.
I updated the dockerfile to preinstall just so use `./img2num --ghcr-img pr-439 sh`

use `just clean` to remove build-wasm and build-c-cpp
use `just build-all` to build all the code and bindings

see the `Justfile` for more - and feel free to add to it

Justfile
Ryan-Millard, line 1: @Krasner please would you make it so it's more like this:

```sh
just cmake build
just docs start
just js build
just wasm build # note that it isn't the same as JS
just py build
just react-js start # use the dev script
just console-cpp build
```

To do that, you'll probably want to set up more advanced files. See [this conversation](https://chatgpt.com/share/6a3170f5-1e70-83ea-b1bb-c194545773a2). I suggest making a folder in the root (named `just`).

Please also try to use the fastest option when building (like `--parallel` for `cmake --build`).

Justfile
```diff
@@ -0,0 +1,36 @@
...
+    rm -rf build-wasm/
+    rm -rf build-c-cpp/
+
+docs:
+    @echo "Generate docusaurus server"
+    pnpm -F docs run build
+
+react-app: build-wasm
+    @echo "Run react sample app"
+    cd example-apps/react-js && pnpm run dev
```
Ryan-Millard:
```suggestion
    pnpm -F react-example run dev
```

Use pnpm filtering to make it work from the root.

Justfile
```diff
@@ -0,0 +1,36 @@
...
+    @echo "Build python bindings and py package"
+    uv sync --reinstall
+    uv pip install opencv-python
+
+build-all: build-c-cpp build-wasm build-py
+
+clean:
+    @echo "Remove build folders"
+    rm -rf build-wasm/
+    rm -rf build-c-cpp/
```
Ryan-Millard: Please define a clean command for each individual thing:

```sh
just js clean
just cpp clean
just docs clean
# and so on
```

this will make it easy to implement a single clean command for all of them.

Dockerfile.dev
```diff
@@ -143,6 +143,11 @@ RUN --mount=type=cache,target=/pnpm/store \
     pnpm install --no-frozen-lockfile
 
 # --------------------------------------------------------------------------------------------------------------
+# Just package manager
+RUN curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh -o setup_just.sh && \
+    bash setup_just.sh --to /usr/local/bin && \
+    rm setup_just.sh
+# --------------------------------------------------------------------------------------------------------------
```
Ryan-Millard, line 150: Please pin this to a specific version. It's unlikely that they will suddenly break things, but it sucks to debug it when they do.


Please also add this so we don't have to be in the root of the project to use just:
```Dockerfile
# Global justfile support
ENV JUSTFILE=/usr/src/app/Justfile
```

Ryan-Millard (changes requested):
@Krasner thank you! This is looking good, but please would you implement the suggestions I made. Most importantly, please would you add more commands.

Dockerfile.dev
Krasner: you can already do that. this isn't needed

Krasner:
`just help` will output a help message:
```
List of commands:
 init: pull submodules
 format: format all files
 build <target>:
         cpp: build c++ core and c bindings
         js: build js/wasm bindings
         py: build python bindings and python package
         all: build all of above
 clean <target>:
         cpp: delete c++ and c build folder (build-c-cpp)
         js: delete js build folder (build-wasm)
 docs <action>:
         build: build docusaurus server
         start: start docusaurus server on port 3000
 react-js <action>:
         build: build example browser app
         start: start example browser app on port 5173
 console-cpp <input_image>: run example C++ app on input image
 console-c <input_image>: run example C app on input image
 console-py <input_image>: run example python app on input image
```

Ryan-Millard:
Hi @Krasner. The changes you made look good but I'm just a bit concerned about scalability and maintainability - specifically with the monolithic file and help argument.

1. The `Justfile` may become very large (I hate files that are longer that 500 lines because they're difficult to navigate and often signal something that could be broken down further). OpenCV does that and I would rather cut my eyes out than read their code and try to understand it because it's so long (look [here](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/contours.cpp) for reference) - I can't even find the Suzuki-Abe implementation.😭
2. The help script will get long and become a major drag in the future. We need a better setup than that because people will forget about it as time progresses, causing us the same type of problem we had in #433. I do have a nice solution in the pnpm setup (run `pnpm run help` to see it), but that's for pnpm, not for Just. I think we should look online at possible solutions.

So far your stuff is great, though. Thanks!

Krasner:
@Ryan-Millard do you want to merge this in it's current state? We can update the Justfile as we go in future PRs but this makes the basic stuff much more accessible already

Ryan-Millard:
> @Ryan-Millard do you want to merge this in it's current state? We can update the Justfile as we go in future PRs but this makes the basic stuff much more accessible already

That sounds like a good idea. We should update the documentation to use the commands in the `Justfile` soon.

@coderabbitai please create 2 issues:

1. docs: update cli commands to use Just tasks added in #439
    - This must include everything in the [_Setup & Dependencies_ page](https://ryan-millard.github.io/Img2Num/info/docs/next/contributing/setup-and-dependencies/)
    - Use #441 as reference because it raises some very good points about the onboarding experience
2. chore(Justfile): refactor to be split based on build system, package manager, and folder
    - Specifically look at the review suggestions I made here - we won't implement them in this PR (because we want Just to be rolled out right away), but they are still relevant in the project
    - Basically just split the Justfile into multiple smaller files to make it easier to manage
    - Make each file have a help script that conforms to the same standard (consider something like what we have for `pnpm run help`, see `scripts/help.js` for an understanding of how the fuzzy finder works). This helper doesn't need to be as complex as the pnpm help script, but it would be nice.

Krasner:
Yes will make a documentation PR after this

Ryan-Millard:
@Krasner please see my most recent commit. If you're happy with it, please will you merge it. I'm happy with this PR.

PR #440, opened 2026-06-18
fix: use length-aware string conversion for wgpu::StringView error logging

### LINKED ISSUE #401: fix(core): use length-aware string conversion for wgpu::StringView error logging (WebGPU)
> [!CAUTION]
> This issue has been claimed by **@laxitajain**, so it is not recommended that you work on it.
>
> > **For @laxitajain:**

---

## Summary

In `core/include/internal/gpu.h`, the device request failure callback streams `msg.data` directly:

```cpp
} else {
    std::cerr << "Device Failed: " << msg.data << std::endl;
}
```

`wgpu::StringView` (and the underlying `WGPUStringView`) is **not** guaranteed to be null-terminated. The API always provides an explicit `length` field, and null termination is only implied when the special sentinel value `WGPU_STRLEN` is used. Streaming `msg.data` directly via `operator<<` (which assumes a null-terminated C string) is therefore unsafe and may read past the end of the buffer or truncate valid message content.

## Expected Fix

Replace the direct `msg.data` reference with a length-aware conversion, for example:

```cpp
} else {
    std::cerr << "Device Failed: "
              << (msg.data && msg.length > 0 ? std::string(msg.data, msg.length) : "Unknown error")
              << std::endl;
}
```

Or use `std::cerr.write(msg.data, msg.length)` where appropriate.

The same pattern is already applied correctly for other `wgpu::StringView` usages in the same file (e.g., the `SetUncapturedErrorCallback` and `SetDeviceLostCallback` handlers).

## Related

- Flagged in PR #396 (comment: https://github.com/Ryan-Millard/Img2Num/pull/396#discussion_r3314505958)
- Related to #240

## Reported by

@Ryan-Millard

laxitajain:
## What was changed & why

Fixes: #401 as suggested

- used std::string(msg.data, msg.length) for length-aware conversion, replacing  direct streaming of wgpu::StringView::data in the device request failure callback

### DIFF
--- a/core/include/internal/gpu.h
+++ b/core/include/internal/gpu.h
@@ -229,7 +229,9 @@ class GPU {
                     device = std::move(d);
                     std::cout << "Device Acquired" << std::endl;
                 } else {
-                    std::cerr << "Device Failed: " << msg.data << std::endl;
+                    std::cerr << "Device Failed: "
+                    << (msg.data && msg.length > 0 ? std::string(msg.data, msg.length) : "Unknown error")
+                    << std::endl;
                 }
                 device_ready = true; // Unblock the loop
             }


Ryan-Millard (approved):
Hey. Thanks for the fix!

It's nice to see you back here again. :)

PR #445, opened 2026-06-19
fix(core): add guard against invalid num_thresholds, preventing division by zero

### LINKED ISSUE #402: fix(core): guard against invalid `num_thresholds` in `threshold_image` (potential divide by >= 0)
> [!CAUTION]
> This issue has been claimed by **@laxitajain**, so it is not recommended that you work on it.
>
> > **For @laxitajain:**

---

## Summary

In `core/src/internal/image_utils.cpp`, the `threshold_image` function computes:

```cpp
const uint8_t REGION_SIZE(255 / num_thresholds);
```

If `num_thresholds` is `0` or negative, this causes undefined behaviour (integer divide-by-zero). Additionally, a `nullptr` `ptr` or non-positive `width`/`height` could lead to further issues downstream.

An early guard should be added before the division to reject invalid inputs.

## Steps to Reproduce

Call `threshold_image` with `num_thresholds == 0` (or any value ≤ 0).

## Expected Behaviour

The function returns early (or signals an error) without performing any division.

## Related

- Flagged in PR #396 (comment: https://github.com/Ryan-Millard/Img2Num/pull/396#discussion_r3314505986)
- Requested by @Ryan-Millard

laxitajain:
## What was changed & why

Added a check as suggested for early return in case of invalid num_thresholds

Fixes: #402


Ryan-Millard:
Hey @laxitajain! Thank you for your work on this. 

Krasner and I have been swamped with work here recently and these kinds of PRs are great help.

One quick note: we changed the branch that PRs merge into - we now merge into `dev` before main so we can test and ensure that everything works.

I'll merge this when the workflows pass. Thank you!🦔🦔

laxitajain:
> Hey @laxitajain! Thank you for your work on this.
> 
> Krasner and I have been swamped with work here recently and these kinds of PRs are great help.
> 
> One quick note: we changed the branch that PRs merge into - we now merge into `dev` before main so we can test and ensure that everything works.
> 
> I'll merge this when the workflows pass. Thank you!🦔🦔

Thanks for the heads up about the dev-branch!

It looks like one of the CI jobs failed while building the React example app. It appears unrelated to this change and is occurring during contributor credits generation:

GitHub API error: 403 while generating contributor credits
vite-plugin-sitemap then fails because dist/robots.txt is missing

Ryan-Millard:
> > Hey @laxitajain! Thank you for your work on this.
> > 
> > Krasner and I have been swamped with work here recently and these kinds of PRs are great help.
> > 
> > One quick note: we changed the branch that PRs merge into - we now merge into `dev` before main so we can test and ensure that everything works.
> > 
> > I'll merge this when the workflows pass. Thank you!🦔🦔
> 
> Thanks for the heads up about the dev-branch!
> 
> It looks like one of the CI jobs failed while building the React example app. It appears unrelated to this change and is occurring during contributor credits generation:
> 
> GitHub API error: 403 while generating contributor credits
> vite-plugin-sitemap then fails because dist/robots.txt is missing

Don't worry about that. It's not your fault. It's either something I broke or a GitHub fault. I'll look into it and sort it out. :)

Ryan-Millard:
@laxitajain I just realised that `black_threshold_image` also needs the same thing. Please would you add the exact same check there as well. Once that is done, I'll merge it.

I'm sorry for the late realisation.

laxitajain:
@Ryan-Millard done.

That is not a problem at all! :)

Ryan-Millard (approved):
Thank you!

PR #504, opened 2026-07-07
docs: improve table of contents

Ryan-Millard:
### Changes

Overhaul docs and styles.

---

@Prachi-Gupta2808 please will you give your input here. Does the Table of Contents actually look better or have I ruined it?😥

We need to completely change the way the site looks so people can distinguish our brand from the others, but I have no clue how to deal with Docusaurus and have been tired of the framework since I started with it.😂😂


Prachi-Gupta2808:
I think the changes are good to go. Table of contents look better too. Are you also thinking of changing how the react live demo one site looks?

Ryan-Millard:
> I think the changes are good to go. Table of contents look better too. Are you also thinking of changing how the react live demo one site looks?

I do think it needs a repaint, but the Docusaurus site should get one that because it's our primary interface.

Issue #527, opened 2026-07-19
Build benchmark page comparing against alternatives


Ryan-Millard:
> [!CAUTION]
> This issue has been claimed by **@krishnasmv**, so it is not recommended that you work on it.
>
> > **For @krishnasmv:**
> > If you would like to revoke your claim, comment `/untake`

---

## Description
Create a comprehensive benchmark page comparing our core pipeline and packages against established alternatives: potrace, vtracer, imagetracerjs, and autotrace. This is critical for differentiating our offering.

## Acceptance Criteria
- [ ] Benchmark page created and accessible
- [ ] Comparisons include speed metrics for core pipeline
- [ ] Comparisons include output quality metrics
- [ ] Offline/browser execution advantage is clearly called out
- [ ] Test data and methodology are documented
- [ ] Results are reproducible
- [ ] Page includes visual comparisons where applicable

## Metrics to Include
- Processing time (for various image sizes)
- Output file size
- Memory usage
- Quality metrics (accuracy, detail preservation)
- Supported platforms and browsers
- Dependencies and requirements

## Implementation Notes
- The offline/browser story is a **high priority** for messaging - make this the centerpiece
- Document all test conditions and images used
- Consider including both synthetic and real-world test cases
- Include link to source code/methodology for transparency

ankushkhakale:
Is the issue still open?
I'm willing to work upon this issue and contribute to the repo. If the issue is still not marked closed, can you please let me work upon it?

Ryan-Millard:
> Is the issue still open? I'm willing to work upon this issue and contribute to the repo. If the issue is still not marked closed, can you please let me work upon it?

Hi @ankushkhakale. Welcome to Img2Num and thank you for your interest in this issue!

I think @Krasner is working on this in #529. For now, it might be best to wait for his response on whether he is still working on it.

Please let me know if you need assistance with anything else. Have a nice day!🦔🦔

Krasner:
> Is the issue still open? I'm willing to work upon this issue and contribute to the repo. If the issue is still not marked closed, can you please let me work upon it?

Hi actually im away for a few weeks. My latest work is here
https://github.com/Krasner/Img2Num/tree/dev/benchmarking2

Look under the example-apps/html-js benchmarking.html

I wanted to add vtracer as a comparison but am having trouble.
https://github.com/visioncortex/vtracer

Can you try to get it in?

Ryan-Millard:
[...]

PR #529, opened 2026-07-21
basic benchmarking demo

Krasner:
Initial demo comparing img2num to imagetracerjs. Will add more as we go.

```
pnpm -F html-js start
```
then go to `example-apps/html-js/benchmarking`

## Related Issues

Fixes: #527 

## Testing & Verification

https://github.com/user-attachments/assets/0fd025d1-9b9a-43f3-97f9-8b6597ebdf2d

### DIFF
--- a/example-apps/html-js/benchmarking.html
+++ b/example-apps/html-js/benchmarking.html
@@ -1,0 +1,529 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
+  <title>Image Vectorization Benchmark</title>
+  <!--script src="imagetracer_v1.2.6.js"></script-->
+  <!--script src="https://cdn.jsdelivr.net/npm/img2num@0.2.1/dist/browser/img2num.js"></script-->
+  <!--script type="module">
+    import * as rawModule from 'https://cdn.jsdelivr.net/npm/img2num@0.2.1/+esm';
+  
+    // Fall back to the raw module if .default isn't defined
+    const img2num = rawModule.default ?? rawModule;
+    
+    window.img2num = img2num;
+  </script-->
+  <script type="module">
+    import * as img2num from "img2num";
+    import ImageTracer from "imagetracerjs";
+    window.img2num = img2num;
+    window.ImageTracer = ImageTracer;
+  </script>
+
+  <style>
+    body {
+      font-family: system-ui, -apple-system, sans-serif;
+      max-width: 1000px;
+      margin: 2rem auto;
+      padding: 0 1rem;
+      background: #f8f9fa;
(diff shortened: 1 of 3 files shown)


Krasner:
@Ryan-Millard let me know your thoughts

Ryan-Millard:
Hi @Krasner. This is really nice! It's way better than what I suggested.😁

It might actually be a good idea to add this instead into the Docusaurus site instead. I'd prefer it there because it can be styled nicely and easily updated without any worry of future changes that may break it.

Initially, I was suggesting something like [the table in this README.md](https://huggingface.co/starvector/starvector-8b-im2svg), but this is much better. I'd still like to have some static benchmarks, though, because some people won't want to do the benchmarking themselves.

After converting it to an SVG, I think we should compare the outputs quantitatively, too. That could be something like rasterizing the outputs and comparing them to the inputs to get a fault rate or something similar. What do you think about that?

Krasner:
> Hi @Krasner. This is really nice! It's way better than what I suggested.😁
> 
> It might actually be a good idea to add this instead into the Docusaurus site instead. I'd prefer it there because it can be styled nicely and easily updated without any worry of future changes that may break it.
> 
> Initially, I was suggesting something like [the table in this README.md](https://huggingface.co/starvector/starvector-8b-im2svg), but this is much better. I'd still like to have some static benchmarks, though, because some people won't want to do the benchmarking themselves.
> 
> After converting it to an SVG, I think we should compare the outputs quantitatively, too. That could be something like rasterizing the outputs and comparing them to the inputs to get a fault rate or something similar. What do you think about that?

Sounds good. What's the general way to add something like this to the docusaurus site?

Yes, your idea about re-rasterizing the svg and computing fidelity metrics is good. One big advantage we have is no holes. The other approaches either introduce large holes or very blocky/non smooth contours.

Ryan-Millard:
> Sounds good. What's the general way to add something like this to the docusaurus site?

Move your code to either `docs/src/pages` or set up a new documentation plugin instance (see `docs/docusaurus.config.js` and [the documentation for it](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs)).

Currently, we add the html-js app to static because it has its own dependencies and whatnot, but adding this to the documentation site would be much better. I think adding a new docs plugin under the `/benchmarks` route would be best because it will natively be styled by Docusaurus and we can write markdown code for it instead of HTML.

> Yes, your idea about re-rasterizing the svg and computing fidelity metrics is good. One big advantage we have is no holes. The other approaches either introduce large holes or very blocky/non smooth contours.

Maybe make the holes have no alpha and put them on a background that will definitely shine through to bring light to why you should use Img2Num.

Ryan-Millard:
# Update

@Krasner is away for a while and unable to work on this change. Anyone interested in working on it is welcome to do so - just le me know and I'll help you get started.

PR #559, opened 2026-08-16
fix(js): restore static wasm URL for bundler asset detection

Ryan-Millard:
## Summary

Fixes a regression in `0.4.0` where the published browser ES build (`dist/browser/img2num.js`) ships a wasm URL that downstream bundlers cannot statically analyze. Any consumer bundling `img2num` with Vite, webpack 5, Rollup, or Parcel gets a runtime 404 (`GET .../assets/img2num.wasm`) followed by `RuntimeError: Aborted(both async and sync fetching of the wasm failed)`.

## Root cause

Vite's lib mode inlines `new URL("img2num.wasm", import.meta.url)` as a data URL (ignoring `assetsInlineLimit`), so the build previously mangled the expression into a non-static form to defeat that analysis:

```js
new URL(globalThis.__IMG2NUM_WASM_NAME__ ??= "img2num.wasm", import.meta.url)
```

That mangled form shipped in the published output. The same property that defeats our lib-mode inlining also blinds **consumers'** bundlers — the exact literal `new URL("<file>", import.meta.url)` is the pattern they rely on to detect, emit, hash, and rewrite the wasm asset in app builds. With it gone, the URL resolves at runtime against the consumer's chunk path, where no wasm exists.

The standalone IIFE/UMD builds are unaffected (`SINGLE_FILE=1`, wasm embedded), which is why the CDN `<script>` path kept working while bundled consumers broke. Node builds resolve via the glue's `__dirname` path and are also unaffected.

## Fix

`preventWasmInlining` is replaced by a two-phase `wasmUrlPlugin`:

1. **`transform`** — mangle the literal into `new URL(globalThis.__IMG2NUM_WASM_NAME__ || "img2num.wasm", import.meta.url)` so lib-mode asset analysis can't inline it.
2. **`generateBundle`** — after analysis is complete, restore the exact literal in the emitted chunk so consumers' bundlers can see it.

Notes on the implementation:

- The sentinel uses `||` instead of `??=`. Logical assignment is ES2021; with `build.target: "es2020"` it gets transpiled before `generateBundle` runs, the restore match fails silently, and the broken expression ships again. This constraint is documented inline.
- The restore match is regex-based (tolerant of quote/whitespace normalization), not an exact-string compare.
- A guard throws if the browser build's output lacks the literal, so this class of regression fails the build instead of publishing. This check would have caught `0.4.0`.
- `globalThis.__IMG2NUM_WASM_NAME__` still works as a manual URL override for exotic setups.

## Result for consumers

Bundling `img2num` is now zero-config: the bundler detects the literal, emits the wasm into the app's assets with a hash, and rewrites the URL. No `?url` imports, no globals, no entry-file ordering requirements. Plain `<script type="module">` consumers copying `img2num.js` + `img2num.wasm` side by side continue to get correct relative resolution.

## Verification

- `grep -c 'new URL("img2num.wasm", import.meta.url)' packages/js/dist/browser/img2num.js` → `1` (literal present in published output; was `0` on `0.4.0`).
- Rebuilt `example-apps/react-js` with **no app-side changes**: build now emits `assets/img2num-MjQHcrHS.wasm` (550 kB) and the vectorization flow works in the served docs build.
- `esm` example unchanged (sibling `img2num.wasm`, relative resolution); `iife`/`umd` examples unchanged (embedded wasm, nothing to fetch); node builds unchanged.

## Follow-up (post-merge)

- [ ] `npm deprecate img2num@0.4.0 "wasm

### DIFF
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -43,7 +43,8 @@
     "./package.json": "./package.json"
   },
   "files": [
-    "dist"
+    "dist",
+    "CHANGELOG.md"
   ],
   "sideEffects": [
     "**/*.wasm",
--- a/packages/js/vite.config.js
+++ b/packages/js/vite.config.js
@@ -62,25 +62,6 @@ if (!T) {
 }
 
 /**
- * Vite inlines `new URL('x.wasm', import.meta.url)` as a data URL in lib mode
- *  and ignores assetsInlineLimit. Hoisting the literal defeats the static
- *  analysis, leaving a real runtime URL resolution against the sibling file.
- */
-function preventWasmInlining() {
-  return {
-    name: "img2num:prevent-wasm-inlining",
-    enforce: "pre",
-    transform(code, id) {
-      if (!id.includes("build-wasm") || !T.copyWasm) return null;
-      if (!code.includes("import.meta.url")) return null;
-
-      const patched = code.replace(/new URL\((["'])(img2num\.wasm)\1,\s*import\.meta\.url\)/g, 'new URL(globalThis.__IMG2NUM_WASM_NAME__ ??= "$2", import.meta.url)');
-      return patched === code ? null : { code: patched, map: null };
-    },
-  };
-}
-
-/**
  * Copies the emitted .wasm into dist/<target>/ after the bundle is written.
  */
 function copyWasmPlugin() {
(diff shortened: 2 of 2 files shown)


Ryan-Millard:
Planned commit message:

```txt
fix: restore static wasm URL for bundler asset detection

The browser ES build published in 0.4.0 shipped a mangled wasm URL
expression, `new URL(globalThis.__IMG2NUM_WASM_NAME__ ??= "img2num.wasm",
import.meta.url)`, introduced to defeat Vite lib-mode data-URL inlining.
The same non-static form also defeats asset detection in consumers'
bundlers (Vite, webpack 5, Rollup, Parcel), which require the exact
literal `new URL("<file>", import.meta.url)` to emit and rewrite the
wasm. Bundled consumers therefore resolved the URL against their own
chunk path at runtime and hit a 404, aborting wasm instantiation.

Replace preventWasmInlining with a two-phase wasmUrlPlugin:

- transform: mangle the literal into a non-static `||` sentinel so
  lib-mode asset analysis cannot inline it. The sentinel must stay
  within build.target (es2020) syntax; `??=` is ES2021 and gets
  transpiled before generateBundle, silently breaking the round-trip.
- generateBundle: restore a bundler-detectable form via a tolerant
  regex. The restored form is a parenthesized ternary whose false
  branch is the exact literal (bundlers match the expression node, so
  detection is unaffected) and whose true branch preserves
  `globalThis.__IMG2NUM_WASM_NAME__` as a runtime override.
- guard: fail the browser build loudly if the literal is missing from
  the output, preventing this regression class from publishing again.

The plugin is gated to the browser target: node glue resolves via
__dirname and standalone glue via document.currentScript, so the URL
pattern only exists in the ES6 web glue. Standalone IIFE/UMD
(SINGLE_FILE=1, embedded wasm) and node builds were never affected.

Verified: literal and override both present in dist/browser/img2num.js;
react-js example emits a hashed img2num-*.wasm with zero app-side
changes; esm/iife/umd examples and node outputs unchanged.

Fixes bundled-consumer wasm 404 in 0.4.0.

chore: add CHANGELOG.md to the package files whitelist

npm only force-includes package.json/README/LICENSE, so release-please's
changelog was absent from published tarballs.
```

PR #566, opened 2026-08-18
fix(py): switch sdist to explicit inclusion mode and exclude Dawn test corpora

Ryan-Millard:
The default inclusion mode leaves include/exclude collisions unspecified: sdist.include "third_party/dawn/**" overlapped every Dawn exclude pattern, producing inconsistent tarballs from the same config (35k vs 93k entries across otherwise identical builds). Explicit mode documents that exclude is applied after include, making sdist contents deterministic. Explicit mode also ignores .gitignore, so local build directories are excluded here too.

Excludes Dawn's test corpora (~60k files never consumed by any build) whose vk-gl-cts tree caused the deterministic sdist unpack failure in CI (see the Build Python failures on #562), the webgpu-cts GN metadata (43 MB test_list.txt plus an 8 MB cache tarball, both also present in the currently published PyPI sdist), the opt-in HermeticXcode macOS toolchain dir that contains the tree's only dangling symlink, and the dangling dawn/.git submodule gitlink.

Result: 105 MB / 93k entries down to 6.5 MB / ~5k entries. The wheel builds from the sdist end-to-end and the console example runs on the GPU path.

inclusion-mode = "explicit" requires scikit-build-core >= 1.0, so the build-system floor is bumped accordingly.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "img2num"
-dynamic = ["version"] # see tool.scikit-buil
+dynamic = ["version"] # derived via [[tool.dynamic-metadata]] below
 description = "A fast and accurate raster image to SVG conversion library"
 requires-python = ">=3.10"
 license = { text = "MIT" }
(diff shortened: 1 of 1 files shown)


Ryan-Millard:
@coderabbitai review

PR #589, opened 2026-09-01
fix: prevent int overflow in Graph::analyzeJunctions vector size calculation

myy04:
### Changes
Added `static_cast<size_t>` to one of the integer arguments of an expression

### Reason
To prevent integer overflow before vector (`junction_map`) construction.

## Related Issues

Fixes: #570 

## Testing & Verification

The project was built and ran with no issues.

PR #590, opened 2026-09-04
refactor(js): make TARGET lookup a strict allowlist

### LINKED ISSUE #580: refactor(js): make TARGET lookup in vite.config.js a strict allowlist
> [!CAUTION]
> This issue has been claimed by **@ShubhamB74**, so it is not recommended that you work on it.
>
> > **For @ShubhamB74:**

---

## Summary

`packages/js/vite.config.js` selects a build target from the `TARGET` environment variable and looks it up in a `TARGETS` table. The lookup is not a strict allowlist, and the target name leaks into an output path. This is not a security problem (the build is run by developers, not untrusted users), but it trips a CodeQL `js/path-injection` alert and has a small correctness wart. A tidy refactor fixes both.

## Current State

```js
const TARGET = process.env.TARGET ?? "browser";
const TARGETS = { browser: {...}, standalone: {...}, "node-esm": {...}, "node-cjs": {...} };

const T = TARGETS[TARGET];
if (!T) {
  throw new Error(`[img2num] Unknown TARGET "${TARGET}". ...`);
}
// ...
const outDir = T.outDir ?? `dist/${TARGET}`;
```

Two issues:

1. **`TARGETS[TARGET]` accepts prototype keys.** `TARGET=constructor` or `TARGET=__proto__` returns something truthy from `Object.prototype`, passes the `if (!T)` guard, and only fails later with a confusing `Missing .../build-wasm/undefined` error instead of the intended "Unknown TARGET" message.
2. **`TARGET` flows into a filesystem path.** Two of the four targets derive `outDir` from the raw string. CodeQL flags this. The other two targets already declare `outDir` explicitly, so the derived form is the odd one out.

## Proposed change

- Add an explicit `outDir` to the `browser` and `standalone` entries (`"dist/browser"` and `"dist/standalone"`, matching current output).
- Replace `const T = TARGETS[TARGET]; if (!T)` with an `Object.hasOwn(TARGETS, TARGET)` check before the lookup.
- Change `const outDir = T.outDir ?? dist/${TARGET}` to `const outDir = T.outDir`.

After this, `TARGET` is only ever used as a table key and every path is built from string literals in `TARGETS`.

## Acceptance criteria

- All four builds produce output in the same directories as before (`dist/browser`, `dist/standalone`, `dist/node`).
- `TARGET=constructor pnpm -F img2num build` fails with the "Unknown TARGET" error, not a missing-directory error, and lists the supported values for TARGET.
- `TARGET=` unset still defaults to `browser`.
- The CodeQL alert no longer reappears on the PR.
- No behaviour change for consumers; this is config-only.

## Notes

- Commit message: `refactor(js): make TARGET lookup a strict allowlist` (conventional commits; the scope is cosmetic, release-please routes by path).
- Target the `dev` branch, not `main`.
- If you're new to the repo, our [contributing guide](https://img2num.dev/docs/contributing/) covers the dev container setup. The wasm build needs Emscripten, which the container provides.

ShubhamB74:
Closes #580

Replaces the `TARGETS[TARGET]` lookup with `Object.hasOwn(TARGETS, TARGET)`
so prototype keys (e.g. TARGET=constructor) are rejected with the intended
"Unknown TARGET" error instead of falling through with a confusing
missing-directory error. Also makes `outDir` explicit for the `browser`
and `standalone` targets so TARGET is never used to build a filesystem
path — resolves the CodeQL js/path-injection alert.

Verified locally:
- TARGET=constructor → "Unknown TARGET" error, as expected
- TARGET=browser (default path) → fails at the pre-existing wasm-missing
  guard, unchanged behavior; full build not run locally since I didn't
  set up the Emscripten dev container for this small change

### DIFF
--- a/packages/js/vite.config.js
+++ b/packages/js/vite.config.js
@@ -14,6 +14,7 @@ const TARGETS = {
     isNode: false,
     // Copy the external .wasm next to the bundle.
     copyWasm: "required",
+    outDir: "dist/browser",
   },
 
   standalone: {
@@ -27,6 +28,7 @@ const TARGETS = {
     isNode: false,
     // Absent when SINGLE_FILE=1, present when SINGLE_FILE=0
     copyWasm: "optional",
+    outDir: "dist/standalone",
   },
 
   "node-esm": {
@@ -56,16 +58,16 @@ const TARGETS = {
   },
 };
 
-const T = TARGETS[TARGET];
-if (!T) {
+if (!Object.hasOwn(TARGETS, TARGET)) {
   throw new Error(`[img2num] Unknown TARGET "${TARGET}". Expected one of: ${Object.keys(TARGETS).join(", ")}`);
 }
+const T = TARGETS[TARGET];
 
 // Single sources of truth for the paths that used to be re-derived in each
 // plugin: glue input dir, output dir, and the remediation hint shown by
 // every "wasm build missing" error.
 const glueDir = path.join(here, "build-wasm", T.glue);
-const outDir = T.outDir ?? `dist/${TARGET}`;
+const outDir = T.outDir;
 const WASM_BUILD_HINT = "Run the CMake wasm build first (just build js, or emcmake cmake -B build-wasm && cmake --build build-wasm).";
 
 // Config-time guard: a missing build-wasm/ otherwise dies later as a cryptic


ShubhamB74:
@Ryan-Millard can you please check this?

Ryan-Millard (approved):
Hi @ShubhamB74, thank you for your contribution and welcome to Img2Num! :)

This is great. I'll merge it once all the workflow pass.

Ryan-Millard:
There's something going on out there in the aether that isn't really your fault - #589 also had some strange problem with the linting workflow and it passes on my computer. We can just ignore it for now.

Thanks again for the PR! Have a nice day.🦔🦔🦔🦔

ShubhamB74:
Thank you so much, Ryan! I’m really glad I could contribute to Img2Num.
I’m looking forward to contributing more and helping the project grow! Have a nice day, you too!

PR #598, opened 2026-09-07
docs(website): add contributor count to homepage stats bar

### LINKED ISSUE #595: docs(website): add contributor count to homepage stats bar
> [!CAUTION]
> This issue has been claimed by **@afshan-khan549**, so it is not recommended that you work on it.
>
> > **For @afshan-khan549:**

---

## Summary

The homepage hero (`docs/src/pages/index.jsx`) shows stars and forks fetched from the GitHub API, but not contributor count. Add a third stat showing the number of contributors.

## Implementation notes

The GitHub API does not have a field that directly gives you the contributor count. There is a simple trick to get it in one request:

1. Request the contributors list, but ask for only **one contributor per page**:

   ```
   GET https://api.github.com/repos/Ryan-Millard/Img2Num/contributors?per_page=1
   ```

2. Look at the `Link` **response header** (not the response body). It tells you where the last page is:

   ```
   link: <...contributors?per_page=1&page=2>; rel="next",
         <...contributors?per_page=1&page=25>; rel="last"
   ```

3. Since each page holds exactly one contributor, the page number in the `rel="last"` entry **is** the total contributor count. In the example above, the count is 25.

This approach works the same whether the repo has 25 contributors or 25,000. No pagination is ever needed.

### Architecture note (read this first)

The build-time counterpart to this issue (#596) will fetch these same stats during the build, as a **Docusaurus plugin**. A plugin is the idiomatic Docusaurus way to load data at build time: it fetches in its `loadContent()` hook, publishes the result with `setGlobalData()`, and components read it back with the `usePluginData()` hook. Do not put the fetching in `docusaurus.config.js` itself via `customFields`; that mixes data loading into site configuration and cannot be reused or tested on its own.

What that means for **this** issue: the fetch logic must not be written inline in the page component. Put it in a shared module so the plugin and the browser can both import the exact same function, with no duplication and no future refactor when the plugin lands:

```
docs/
  lib/
    githubStats.js      <- shared fetch logic lives here (this issue creates it)
  plugins/
    github-stats/       <- build-time plugin imports from lib/ (other issue)
  src/
    pages/index.jsx     <- browser code imports from lib/ (this issue)
```

The function below uses only the standard `fetch()` API, which works identically in the browser and in Node 18+, so the same file runs in both places without changes.

### Step 1: Create the shared helper module

Create `docs/lib/githubStats.js`. It does the whole trick described above and returns a plain number, or `null` if anything goes wrong:

```js
const REPO = "Ryan-Millard/Img2Num";

// Works in the browser and at build time (Node 18+).
// The optional token is only used at build time, where CI provides one.
export async function fetchContributorCount({ token } = {}) {
  try {
    const res = await fetch(
      `https://api.github.com/repos/${REPO}/contributors?per_page=1`,
      {
        headers: token ? { Authorization: `Bearer ${token}` } : {},
      }
    );

    // Rate limits, network problems, and so on. Error responses have no
    // Link header, so without this check a failure could look like a
    // low contributor count.
    if (!res.ok) return null;

    const link = res.headers.get("link");

    if (link) {
      // Grab the page number from the rel="last" entry.
      const match = link.match(/[?&]page=(\d+)>;\s*rel="last"/);
      return match ? Number(match[1]) : null;
    }

    // OK response but no Link header: everything fits on one page,
    // so the count is just the length of the returned array.
    const body = await res.json();
    return body.length;
  } catch {
    return null;
  }
}
```

The `token` option costs nothing now and means the plugin issue can reuse this function as-is, without touching this file. Never pass a token from browser code; it would be visible to every visitor.

In `docs/src/pages/index.jsx`, add `contributors` to the existing state and fetch it in the same `useEffect` that already fetches stars and forks:

```js
const [stats, setStats] = useState({
  stars: null,
  forks: null,
  contributors: null,
});

useEffect(() => {
  fetch("https://api.github.com/repos/Ryan-Millard/Img2Num")
    .then((res) => (res.ok ? res.json() : null))
    .then((data) => {
      if (data) {
        setStats((prev) => ({
          ...prev,
          stars: data.stargazers_count,
          forks: data.forks_count,
        }));
      }
    })
    .catch(() => {});

  fetchContributorCount().then((count) => {
    if (count !== null) {
      setStats((prev) => ({ ...prev, contributors: count }));
    }
  });
}, []);
```

Note the use of `setStats((prev) => ...)` instead of `setStats({...})`. The two fetches finish at different times, and this form makes sure the slower one does not wipe out the result of the faster one.

### Step 3: Show it in the stats bar

Add a third stat next to the existing two:

```jsx
<div className={styles.stat}>
  <span className={styles.statNum}>
    ◉ {stats.contributors !== null ? stats.contributors : "—"}
  </span>
  <span className={styles.statLabel}>contributors</span>
</div>
```

The `◉` glyph is a placeholder. Pick whatever matches the existing ★ and ⑂ style.

### Expected behaviour (not bugs)

* The count **includes bots** such as `dependabot[bot]` and `github-actions[bot]`. The `Link` header only gives a number, so bots cannot be filtered out.
* The count will be **lower than the number shown on the github.com repo page**. The API counts distinct commit authors on the default branch. The website additionally credits people listed in `Co-authored-by` commit trailers. The website's number is not available through any API.

## Requirements

* Stats bar shows a `contributors` stat next to `stars` and `forks`
* Count is read from the `Link` header (no full contributor list download, no pagination)
* Fetch failure degrades gracefully: show `—`, no console errors, and never display a wrong low number
* Glyph and style consistent with the existing ★ / ⑂ stats
* Responsive design (especially small screens, where there is little space)

## Related

Related to #596, which will seed all GitHub stats at build time instead of client-only fetching.

* If **this issue lands first**, that issue must include contributors in the build-time seeding.
* If **that issue lands first**, this one must follow its pattern: add contributors to the build-time fetch in `docusaurus.config.js` **and** the client-side refresh, not client-only.

afshanos:
Closes #595

## Changes
- Added `docs/lib/githubStats.js` with `fetchContributorCount()`, using the Link-header trick described in the issue
- Added a third stat (contributors) to the homepage stats bar, next to stars and forks

## Verification
- Confirmed the contributors count renders correctly (matches GitHub's contributor count)
- Confirmed graceful failure handling: simulated a failed request and verified the stat falls back to `—` with no console errors, and doesn't display a wrong low number

### DIFF
--- a/docs/lib/githubStats.js
+++ b/docs/lib/githubStats.js
@@ -1,0 +1,25 @@
+const REPO = "Ryan-Millard/Img2Num";
+
+// Works in the browser and at build time (Node 18+).
+// The optional token is only used at build time, where CI provides one.
+export async function fetchContributorCount({ token } = {}) {
+  try {
+    const res = await fetch(`https://api.github.com/repos/${REPO}/contributors?per_page=1`, {
+      headers: token ? { Authorization: `Bearer ${token}` } : {},
+    });
+
+    if (!res.ok) return null;
+
+    const link = res.headers.get("link");
+
+    if (link) {
+      const match = link.match(/[?&]page=(\d+)>;\s*rel="last"/);
+      return match ? Number(match[1]) : null;
+    }
+
+    const body = await res.json();
+    return body.length;
+  } catch {
+    return null;
+  }
+}
--- a/docs/src/pages/index.jsx
+++ b/docs/src/pages/index.jsx
@@ -6,6 +6,7 @@ import React, { useEffect, useState } from "react";
 import Hedgehog from "../components/Hedgehog";
 import styles from "./index.module.css";
 import CodeBlock from "@theme/CodeBlock";
+import { fetchContributorCount } from "../../lib/githubStats";
 
 function RasterToSvgDemo() {
   return (
@@ -42,6 +43,7 @@ function HeroSection() {
   const [stats, setStats] = useState({
     stars: null,
     forks: null,
+    contributors: null,
   });
 
   useEffect(() => {
@@ -49,13 +51,20 @@ function HeroSection() {
       .then((res) => (res.ok ? res.json() : null))
       .then((data) => {
         if (data) {
-          setStats({
+          setStats((prev) => ({
+            ...prev,
             stars: data.stargazers_count,
             forks: data.forks_count,
-          });
+          }));
         }
       })
       .catch(() => {});
+
+    fetchContributorCount().then((count) => {
+      if (count !== null) {
+        setStats((prev) => ({ ...prev, contributors: count }));
+      }
+    });
   }, []);
 
   return (
@@ -106,6 +115,10 @@ function HeroSection() {
             <div className={styles.stat}>
               <span className={styles.statNum}>⑂ {stats.forks !== null ? stats.forks : "—"}</span>
               <span className={styles.statLabel}>forks</span>
+            </div>
+            <div className={styles.stat}>
+              <span className={styles.statNum}>◉ {stats.contributors !== null ? stats.contributors : "—"}</span>
+              <span className={styles.statLabel}>contributors</span>
             </div>
             <div className={styles.stat}>
               <span className={styles.statNum}>C++/C/Py/JS</span>


Ryan-Millard (approved):
Hi @afshanos. Thank for this - it looks great! :)

PR #597, opened 2026-09-07
docs(site): add client-side redirect from /docs/faq to /faq

afshanos:
Closes #591

## Changes
- Added `@docusaurus/plugin-client-redirects` as a dependency
- Configured a client-side redirect from `/docs/faq` to `/faq` in `docs/docusaurus.config.js`

## Verification
Since the plugin only generates redirects at build time, I verified with a production build rather than `pnpm run dev`:

pnpm run build && pnpm run serve

Confirmed `/docs/faq` redirects to `/faq` correctly in the browser.

## Note
`dev` currently fails a full production build due to a pre-existing, unrelated broken link:

Broken link on source page path = /docs/py/:
-> linking to ./api-reference/ (resolved as: /docs/py/api-reference/)

I confirmed this happens even on a clean `dev` checkout with none of my changes, so it's unrelated to this PR. To verify my redirect end-to-end, I temporarily set `onBrokenLinks: "warn"` locally - that change is **not** included in this PR. 
Happy to open a separate issue for the broken link if useful.

PR #605, opened 2026-09-17
build(website): seed GitHub stats at build time to avoid unpopulated first paint

### LINKED ISSUE #596: build(website): Seed GitHub stats at build time to avoid unpopulated first paint
> [!CAUTION]
> This issue has been claimed by **@afshan-khan549**, so it is not recommended that you work on it.
>
> > **For @afshan-khan549:**

---

# Seed GitHub stats at build time to avoid unpopulated first paint

## Summary

The homepage stats bar (`docs/src/pages/index.jsx`) is client-fetch only, so every visitor briefly sees `—` placeholders until the GitHub API responds, or permanently if the request fails (rate limit, blocked, offline). Fetch the stats at build time so real numbers are in the static HTML, then let the existing client-side fetch refresh them.

## Approach

Implement the build-time fetching as a **Docusaurus plugin**. This is the idiomatic Docusaurus way to load data during the build:

1. The plugin fetches the stats in its `loadContent()` hook, which Docusaurus runs at build time (in Node).
2. It publishes the result with `setGlobalData()`.
3. The page component reads it back with the `usePluginData()` hook and uses it as the initial state.
4. The existing client-side fetch stays and refreshes the numbers after the page loads.

Do **not** fetch in `docusaurus.config.js` via `customFields`. That works, but it mixes data loading into site configuration, cannot be tested on its own, and does not scale if we ever seed more data.

All fetch logic must live in the shared module `docs/lib/githubStats.js` so the plugin and the browser use the exact same functions. That module is created by #595; see the Related section at the bottom for how the two issues coordinate.

```
docs/
  lib/
    githubStats.js      <- shared fetch logic (browser + Node)
  plugins/
    github-stats/
      index.js          <- the plugin (this issue creates it)
  src/
    pages/index.jsx     <- reads plugin data, keeps client refresh
```

### Step 1: Add a repo stats function to the shared module

In `docs/lib/githubStats.js`, next to the existing `fetchContributorCount()`, add a function for stars and forks in the same style: it returns the data, or `null` if anything at all goes wrong.

```js
const REPO = "Ryan-Millard/Img2Num";

// Works in the browser and at build time (Node 18+).
// The optional token is only used at build time, where CI provides one.
export async function fetchRepoStats({ token } = {}) {
  try {
    const res = await fetch(`https://api.github.com/repos/${REPO}`, {
      headers: token ? { Authorization: `Bearer ${token}` } : {},
    });
    if (!res.ok) return null;

    const data = await res.json();
    return {
      stars: data.stargazers_count,
      forks: data.forks_count,
    };
  } catch {
    return null;
  }
}
```

### Step 2: Create the plugin

Create `docs/plugins/github-stats/index.js`:

```js
import { fetchContributorCount, fetchRepoStats } from "../../lib/githubStats.js";

export default function githubStatsPlugin() {
  return {
    name: "github-stats",

    // Runs at build time, in Node.
    async loadContent() {
      const token = process.env.GITHUB_TOKEN;

      const [repo, contributors] = await Promise.all([
        fetchRepoStats({ token }),
        fetchContributorCount({ token }),
      ]);

      // Every field falls back to null. The shared functions never throw,
      // so a GitHub outage or rate limit can never fail the build.
      return {
        stars: repo?.stars ?? null,
        forks: repo?.forks ?? null,
        contributors: contributors ?? null,
      };
    },

    // Makes the loaded content available to components.
    async contentLoaded({ content, actions }) {
      actions.setGlobalData(content);
    },
  };
}
```

Register it in `docusaurus.config.js`:

```js
plugins: [
  "./plugins/github-stats/index.js",
  // Other plugins
],
```

### Step 3: Use the plugin data in `HeroSection`

In `docs/src/pages/index.jsx`, read the build-time numbers with `usePluginData()` and use them as the **initial** state. Keep the client-side fetch, but make sure a failed fetch can never erase a build-time value. The `?? prev.x` pattern does that: if the fresh value is null, keep the previous one.

```js
import { usePluginData } from "@docusaurus/useGlobalData";
import { fetchContributorCount, fetchRepoStats } from "../../lib/githubStats";
```

```js
function HeroSection() {
  // Build-time numbers, baked into the static HTML.
  const buildTimeStats = usePluginData("github-stats");

  const [stats, setStats] = useState(buildTimeStats);

  // Client-side refresh, so numbers stay current between deploys.
  useEffect(() => {
    Promise.all([fetchRepoStats(), fetchContributorCount()]).then(
      ([repo, contributors]) => {
        setStats((prev) => ({
          stars: repo?.stars ?? prev.stars,
          forks: repo?.forks ?? prev.forks,
          contributors: contributors ?? prev.contributors,
        }));
      }
    );
  }, []);

  // The rest of the component stays the same
}
```

Note that the old inline `fetch("https://api.github.com/repos/...")` call in the component gets deleted. After this change, the component contains no fetch logic of its own; everything goes through the shared module.

### Step 4: Give the build a token in CI

Unauthenticated GitHub API requests are limited to 60 per hour **per IP address**. That is fine on your own machine, but CI runners share IP addresses with strangers, so the limit is often already used up (this exact failure was observed while researching this issue). GitHub Actions provides a token automatically; expose it to the build.

The docs build lives in the reusable workflow `.github/workflows/build-docs.yml` (called from `ci.yml` and `deploy.yml`). Add `GITHUB_TOKEN` to the env of the existing "Build Docusaurus" step, exactly like the Algolia variables already there:

```yaml
      - name: Build Docusaurus
        env:
          ALGOLIA_APP_ID: ${{ secrets.ALGOLIA_APP_ID }}
          ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
          ALGOLIA_INDEX_NAME: ${{ secrets.ALGOLIA_INDEX_NAME }}
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: pnpm -F docs run build
```

Two things that make this a one-file change:

* Unlike the Algolia secrets, `GITHUB_TOKEN` must **not** be added to the `secrets:` declarations in the `workflow_call` trigger, and the calling workflows (`ci.yml`, `deploy.yml`) need no changes. `GITHUB_TOKEN` is automatically available to reusable workflows, and its name is reserved, so declaring it as a custom secret would actually fail.
* The workflow's `permissions: contents: read` is already enough. Reading public repo stats needs no extra permissions.

Local builds without a token keep working (`pnpm -F docs run build` on your machine); they just use the unauthenticated limit of your own IP.

## Acceptance criteria

* Stats fetching is implemented as a Docusaurus plugin (`loadContent` + `setGlobalData`), not via `customFields`
* All fetch logic lives in `docs/lib/githubStats.js` and is shared between the plugin and the browser; the page component contains no fetch calls of its own
* Built HTML contains real star and fork counts (verify by loading the page with JavaScript disabled)
* Client-side fetch still updates the numbers after page load
* A failed client fetch leaves build-time values intact (no regression to `—`)
* The build succeeds even when the GitHub API is unreachable (stats fall back to null, page shows `—`)
* The "Build Docusaurus" step in `.github/workflows/build-docs.yml` passes `GITHUB_TOKEN`, with no changes needed in `ci.yml` or `deploy.yml`

## Related

Related to #595, which adds a contributor count stat and creates the shared `docs/lib/githubStats.js` module.

* If **this issue lands first**, create the shared module here with both functions, and seed all three stats (stars, forks, contributors). #595 then only adds the stat to the UI.
* If **#595 lands first**, the shared module and `fetchContributorCount()` already exist. This issue adds `fetchRepoStats()` next to it and wires both into the plugin.

afshanos:
Closes #596

## Changes
- Added `fetchRepoStats()` to `docs/lib/githubStats.js`, next to the existing `fetchContributorCount()`
- Created `docs/plugins/github-stats/index.js` as a Docusaurus plugin, using `loadContent()` + `setGlobalData()` to fetch stats at build time
- Registered the plugin in `docs/docusaurus.config.js`
- Updated `HeroSection` in `docs/src/pages/index.jsx` to read build-time stats via `usePluginData()` as initial state, with the client-side fetch now only refreshing (never erasing) those values
- Added `GITHUB_TOKEN` to the "Build Docusaurus" step's env in `.github/workflows/build-docs.yml`, so CI builds use an authenticated request and avoid rate limiting

## Verification
- Ran a production build (`docusaurus build`) and confirmed the plugin's `loadContent()` executes successfully with no errors
- Served the build and loaded the homepage with **JavaScript disabled** - confirmed real stats (★ 65, ⑂ 47, ◉ 31) render directly from the static HTML, proving build-time seeding works
- Re-enabled JavaScript and confirmed the client-side refresh still runs and updates the numbers
- Verified the component contains no inline fetch calls of its own- all fetching goes through the shared module

## Note
Same as #597 and #598: `dev` currently has a pre-existing, unrelated broken link at `/docs/py/` that causes `onBrokenLinks: "throw"` to fail full production builds. I temporarily set it to `"warn"` locally to complete verification- that change is **not** included in this PR.


Ryan-Millard (approved):
Great stuff. Thank you!

We can ignore the linting warning because the linter is broken. It'll be fixed relatively soon.

